PR #29, opened 2018-04-27
Add STDIN and STDOUT support

alok:
If `-` passed on command line, output to STDOUT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ Sample `.pre-commit-config.yaml`:
     -   id: pyupgrade
 ```
 
+## As a UNIX filter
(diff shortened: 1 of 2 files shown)


asottile:
This doesn't pass tests -- also if we're going to support this mode, I'd like to make it consistent with something else, perhaps the `--print-only` option of [reorder-python-imports](https://github.com/asottile/reorder_python_imports/blob/794f0498340ab3adbef443431993f8a8b2046da7/reorder_python_imports.py#L390-L393)?

asottile:
closing (inactivity)

PR #36, opened 2018-06-06
Implement rewriting of simple f-strings

### LINKED ISSUE #28: f-string roadmap?
You mention that f-string support was planned in the README, and I'm just curious if you had a rough idea of when that might be.

Big fan of the project by the way.

asottile:
resolves #28


asottile:
well that was easy

PR #64, opened 2018-10-08
Simplify parse_format

vmarkovtsev:


asottile:
This is intentionally styled this way.  The explicit `else` block makes the code easier to parse for humans because you can notice at a glance that there are two distinct, disparate code paths

vmarkovtsev:
@asottile FWIW it's my turn to give you links: https://stackoverflow.com/questions/9191388/it-is-more-efficient-to-use-if-return-return-or-if-else-return
:smile:

asottile:
In python it matters not at all to the interpreter so optimize for the human:

```pycon
>>> def f(x):
...     if x:
...         return 1
...     else:
...         return 2
... 
>>> def g(x):
...     if x:
...          return 1
...     return 2
... 
>>> dis.dis(f)
  2           0 LOAD_FAST                0 (x)
              2 POP_JUMP_IF_FALSE        8

  3           4 LOAD_CONST               1 (1)
              6 RETURN_VALUE

  5     >>    8 LOAD_CONST               2 (2)
             10 RETURN_VALUE
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE
>>> dis.dis(g)
  2           0 LOAD_FAST                0 (x)
              2 POP_JUMP_IF_FALSE        8

  3           4 LOAD_CONST               1 (1)
              6 RETURN_VALUE

  4     >>    8 LOAD_CONST               2 (2)
             10 RETURN_VALUE
```

PR #67, opened 2018-10-08
Add --keep-percent-format

### LINKED ISSUE #63: Add the ability to disable specific upgrades
E.g. our style guide forces printf-like syntax (we switch to C-like languages often and cut the corners) and we have to revert the corresponding upgrades as in https://github.com/src-d/ml/pull/322#pullrequestreview-162382042

So would be nice to enable/disable features besides the master switch by Python version.

asottile:
Resolves #63

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,9 @@ dict([(a, b) for a, b in y)  # {a: b for a, b in y}
 
 ### printf-style string formatting
 
+Availability:
+- Unless `--keep-percent-format` is passed.
+
 ```python
 '%s %s' % (a, b)                  # '{} {}'.format(a, b)
 '%r %2f' % (a, b)                 # '{!r} {:2f}'.format(a, b)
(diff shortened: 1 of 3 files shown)


asottile:
CC @hugovk @vmarkovtsev 

sorry again <3

hugovk:
Thanks!

asottile:
More context:

I made a mistake here, I was unnecessarily strict on an ideal that isn't entirely realistic.  I don't really want to force / impose my (unrealistically strict and high) expectations on everyone else.  This in the end was a very simple change which enables this tool to be accessible for a lot more people at little cost.  I ate breakfast, felt better, and decided this wasn't so bad.

Also I realized this solves my predicament with blanket applying this in `pytest` where there's a significant amount of %-format left around (and in some cases, _intentionally_ due to the differences with `.format()`).

Apologies again, I'll try and be more reasonable next time

PR #87, opened 2018-12-08
Add testing and document support for all active Python branches

jdufresne:


asottile:
is there a motivation for this?

- `dist: xenial` results in slower travis-ci, if applied at all it should [only be applied to the python3.7 env](https://github.com/deadsnakes/travis-ci-python3.7-example#update-2018-07-03)
- I already test python3.7 on appveyor, testing it again on linux increases ci time for not much benefit
- I have no interest in supporting python3.4, especially if there's no code changes
- same goes for python3.5, but I guess it's there because I haven't gotten around to removing it with [all-repos](https://github.com/asottile/all-repos)
- I have no interest in testing pypy3 unless there's code changes necessary -- it's basically the same as python3.5 and I already test pypy

jdufresne:
ok

hugovk:
Perhaps the 3.7 classifier would still be useful?

And maybe `python_requires` as well for 2.7 and 3.6+?

asottile:
the classifier seems fine, and the tox change seems fine.  I'm still not sold on `python_requires` being useful.  Despite not testing it, pyupgrade _probably_ works fine all the way back to python3.3, maybe even 3.2 if the tests didn't have `u'literals'`

jdufresne:
I've reduced the PR to the following:

- Add 3.7 trove classifier to document support
- Remove 3.5 testing
- Include py37 in `tox.ini`, but not in `.travis.yml`

>  I'm still not sold on python_requires being useful. Despite not testing it, pyupgrade probably works fine all the way back to python3.3, maybe even 3.2 if the tests didn't have u'literals'

IMO, this sends a mixed message to potential users and contributors. If the project doesn't support these older Pythons, it would be really good to let users, scripts, and pip know through trove classifiers and `python_requires`. I often use the trove classifier to help know, at a glance, if a project is suitable as a dependency. Adding `python_requires` helps prevent the package from being installed to unsupported environments. On the other hand, if the project does support older Pythons, it would be good to test them so new contributions don't introduce unwanted breakage. Of course, this decision is up to you. I have not included that in this PR.

asottile:
Thanks!

PR #90, opened 2019-01-14
Include the LICENSE file in the sdist

nicoddemus:
🙃

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [bdist_wheel]
 universal = True
+
+[metadata]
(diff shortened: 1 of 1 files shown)


sloria:
I believe the preferred way to do this is to add 

```ini
[metadata]
license_file = LICENSE
```

to `setup.cfg`.

nicoddemus:
@sloria thanks for the tip, that's indeed better.

Updated. 👍

asottile:
I wonder if we shouldn't just make setuptools do this by default (doesn't it already?)

PR #91, opened 2019-01-15
Test against Python 3.7

sloria:
Also, improve formatting for better vertical alignment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
 language: python
 matrix:
     include:
-        -   env: TOXENV=py27
-        -   env: TOXENV=py36
-            python: 3.6
-        -   env: TOXENV=pypy
-            python: pypy-5.7.1
+        -   { python: '2.7', env: TOXENV=py27 }
+        -   { python: '3.6', env: TOXENV=py36 }
+        -   { python: '3.7', env: TOXENV=py37, dist: xenial, sudo: true }
+        -   { python: 'pypy-5.7.1', env: TOXENV=pypy }
 install: pip install coveralls tox
 script: tox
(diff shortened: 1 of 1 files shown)


asottile:
- [we're already testing against py37 on windows](https://github.com/asottile/pyupgrade/blob/bf9aa8f3fdabcd34542ba1de829759b6addb4bef/appveyor.yml#L4)
- [`sudo: true` is no longer necessary for `dist: xenial`](https://github.com/deadsnakes/travis-ci-python3.7-example#update-2018-07-03)
- I'd rather keep the indentation that's already there, it's at least consistent with my other projects
- I plan to move all my projects to py27 + py37 once I don't have to jump through hoops for travis

so probably won't do this now? and then later I'll automate it with [all-repos](https://github.com/asottile/all-repos) -- so, thanks for the PR 🎉 but probably not today :(

sloria:
Cool cool. TIL `sudo: true` is no longer necessary for `dist: xenial`. 👍

asottile:
<3 thank you though the PR is still appreciated!

hugovk:
`sudo:` is no longer needed, full stop.
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Personally, I would add 3.7 to Travis: if a contributor is going to turn on a CI before submitting, it's more likely to be Travis than AppVeyor. Also, it's still inside the 5 job limit, and who knows when you might find a bug on 3.7 on Linux but not Windows?

asottile:
didn't know they had progressed their migration so quickly

cleaned it from all my repos 🎉 https://github.com/asottile/all-repos-incantations#travis-remove-sudo

PR #96, opened 2019-01-31
Actually include LICENSE in sdist

sloria:
Turns out #90 only adds LICENSE to the wheel.
This ensures that LICENSE and README.md are also
in the sdist.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include *.md LICENSE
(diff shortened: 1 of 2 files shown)


asottile:
I'd rather fix https://github.com/pypa/setuptools/issues/1636 :S

sloria:
Yes, that would be preferable.

asottile:
Turns out setuptools 40.8 now supports the `setup.cfg` metadata without needing `MANIFEST.in`:

https://setuptools.readthedocs.io/en/latest/history.html#v40-8-0

I'll be `all-repos`-ing all my repositories to apply that _soon_

PR #101, opened 2019-02-07
Add py38 to Tox

hugovk:
`# These should match the travis env list`

So remove `py37`?


asottile:
I'm going to wait until 3.8 actually is installable instead of building it out of my homedir.  the comment is kinda copypasta bogus that I'll probably purge eventually anyway

asottile:
I'll handle this in an automated way at a later date -- thanks for the PR!

Issue #59, opened 2018-10-07
six: burn the bridges


asottile:
Let's try and automatically un-six python source.  Here's a quick audit of what six does and how to undo it in python 3:

From [the docs](https://six.readthedocs.io/):

### trivial replacements
- [x] `six.text_type` => `str`
- [x] `six.binary_type` => `bytes`
- [x] `six.class_types` => `(type,)`
- [x] `six.string_types` => `(str,)`
- [x] `six.integer_types` => `(int,)`
- [x] `six.unichr` => `chr`
- [x] `six.iterbytes` => `iter`
- [x] `six.print_(...)` => `print(...)`
- [x] `six.exec_(c, g, l)` => `exec(c, g, l)`
- [x] `six.advance_iterator(it)` => `next(it)`
- [x] `six.next(it)` => `next(it)`
- [x] `six.callable(x)` => `callable(x)`

### trivial removals:
- [x] `python_2_unicode_compatible` (remove decorator)

   ```python
   @six.python_2_unicode_compatible
   class C: pass
   ```

### requires parsing a function call
- [x] `six.u('foo')` => `'foo'`
- [x] `six.byte2int(bs)` => `bs[0]`
- [x] `six.indexbytes(bs, i)` => `bs[i]`
- [x] `six.iteritems(dct)` => `dct.items()  # also {iter,view}{items,keys,values}`
- [x] `six.create_unbound_method(fn, cls)` => `fn`
- [x] `six.get_unbound_method(meth)` => `meth`
- [x] `six.get_method_function(meth)` => `meth.__func__`
- [x] `six.get_method_self(meth)` => `meth.__self__`
- [x] `six.get_function_closure(fn)` => `fn.__closure__`
- [x] `six.get_function_code(fn)` => `fn.__code__`
- [x] `six.get_function_defaults(fn)` => `fn.__defaults__`
- [x] `six.get_function_globals(fn)` => `fn.__globals__`
- [x] `six.assertCountEqual(self, arg1, arg2)` => `self.assertCountEqual(arg1, arg2)`
- [x] `six.assertRaisesRegex(self, fn, *a, **k)` => `self.assertRaisesRegex(fn, *a, **k)`
- [x] `six.assertRegex(self, *a)` => `self.assertRegex(*a)`

### maybe replace with literal?
- [x] `six.b` => replace with bytes literal?

### need to know that you're not in a place where a `raise` statement is invalid
[...]

PR #116, opened 2019-03-18
Rewrite six.b(...) to b-literal

mxr:
Taking a stab at this. Not sure if this is the best approach or if there are edge cases I missed. This is part of #59.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -669,6 +669,13 @@ def _is_bytestring(s):
         return False
 
 
+def _is_ascii(s):
+    if sys.version_info >= (3, 7):  # pragma: no cover (py37+)
+        return s.isascii()
+    else:  # pragma: no cover (<py37)
+        return all(c in string.printable for c in s)
+
+
 def _fix_percent_format_tuple(tokens, start, node):
     # TODO: this is overly timid
     paren = start + 4
@@ -977,6 +984,7 @@ SIX_TYPE_CTX_ATTRS = dict(
 )
 SIX_CALLS = {
     'u': '{args[0]}',
+    'b': 'b{args[0]}',
     'byte2int': '{arg0}[0]',
     'indexbytes': '{args[0]}[{rest}]',
     'iteritems': '{args[0]}.items()',
@@ -1193,7 +1201,10 @@ def _fix_six(contents_text):
                 del tokens[i - 1:end + 1]
         elif token.offset in visitor.call_names:
             node = visitor.call_names[token.offset]
-            if tokens[i + 1].src == '(':
+            if (
+                    tokens[i + 1].src == '(' and
+                    (tokens[i].src != 'b' or _valid_six_b(tokens, i))
+            ):
                 func_args, end = _parse_call_args(tokens, i + 1)
                 template = SIX_CALLS[node.func.id]
                 _replace_call(tokens, i, end, func_args, template)
@@ -1202,7 +1213,8 @@ def _fix_six(contents_text):
             if (
                     tokens[i + 1].src == '.' and
                     tokens[i + 2].src == node.func.attr and
-                    tokens[i + 3].src == '('
+                    tokens[i + 3].src == '(' and
+                    (tokens[i + 2].src != 'b' or _valid_six_b(tokens, i + 2))
             ):
                 func_args, end = _parse_call_args(tokens, i + 3)
                 template = SIX_CALLS[node.func.attr]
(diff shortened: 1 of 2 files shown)


pyupgrade.py
```diff
@@ -1197,6 +1207,9 @@ def _replace_attr(i, mapping, node):
                 func_args, end = _parse_call_args(tokens, i + 1)
                 template = SIX_CALLS[node.func.id]
                 _replace_call(tokens, i, end, func_args, template)
+        elif token.offset in visitor.literal_attrs:
+            func_args, end = _parse_call_args(tokens, i + 3)
+            _replace_call(tokens, i, end, func_args, 'b{args[0]}')
```
asottile: if it's this easy (?) I think this can just be an entry in the `SIX_CALLS` mapping above

asottile: oh except it has to be a literal 🤔

mxr: It seems like using a different set of attrs is a cleaner way to implement this special case

asottile:
looks like the only edge case (that isn't currently handled) is the literal has to be entirely ascii characters (even though `b` is implemented as `.encode('latin-1')`

so `b('\x12\xef')` is fine, but `b('£')` is not safely rewritable (despite being in latin-1)

mxr:
Good point. I'll update this tomorrow

mxr:
Maybe I'm not understanding what you mean by entirely ascii characters? 

```console
$ python3.7 -c "print('\x123\xef'.isascii())"
False
```

Also I realized I have to detect `from six import b`

asottile:
> Maybe I'm not understanding what you mean by entirely ascii characters?
> 
> ```
> $ python3.7 -c "print('\x123\xef'.isascii())"
> False
> ```
> 
> Also I realized I have to detect `from six import b`

Not the string itself, but the source.

Hope this demonstrates:

```pycon
>>> from six import b
>>> b('foo') == b'foo'
True
>>> b('\xa1') == b'\xa1'
True
>>> b('£')
b'\xa3'
>>> b('£') == b'£'
  File "<stdin>", line 1
SyntaxError: bytes can only contain ASCII literal characters.
```

mxr:
@asottile let me know how this looks?

pyupgrade.py
```diff
@@ -669,6 +669,20 @@ def _is_bytestring(s):
...
+        return s.isascii()
+    else:  # pragma: no cover (<py37)
+        try:
+            s.encode('ascii')
+        except UnicodeDecodeError:  # pragma: no cover (py2)
+            return False
+        except UnicodeEncodeError:  # pragma: no cover (py2+)
+            return False
+        else:
+            return True
```
asottile: some of these branches aren't necessary -- the tokens you get will always be text -- did this get copy pasted from somewhere?

pyupgrade.py
```diff
@@ -1086,6 +1101,22 @@ def visit_Call(self, node):
...
+                isinstance(node.args[0], ast.Str)
+        ):
+            self.literal_attrs[_ast_to_offset(node)] = node
+        elif (
+                isinstance(node.func, ast.Name) and
+                node.func.id == 'b' and
+                node.func.id in self.six_from_imports and
+                isinstance(node.args[0], ast.Str)
+        ):
+            self.literal_attrs[_ast_to_offset(node)] = node
```
asottile: any way that `is_six_attr` / `is_six_name` can be reused?

pyupgrade.py
```diff
@@ -1197,6 +1228,20 @@ def _replace_attr(i, mapping, node):
...
+            template = 'b{args[0]}'
+            if tokens[i + 1].src == '(' and _is_ascii(tokens[i + 2].src):
+                func_args, end = _parse_call_args(tokens, i + 1)
+                _replace_call(tokens, i, end, func_args, template)
+            elif (
+                    tokens[i + 1].src == '.' and
+                    tokens[i + 2].src == node.func.attr and
+                    tokens[i + 3].src == '(' and
+                    _is_ascii(tokens[i + 4].src)
+            ):
```
asottile: ugly, but I think this current breaks:

```
six.b(    "eurosymbol")
```

pyupgrade.py
mxr: I think I was playing with b’’ strings in the interpreter. Will revisit this

pyupgrade.py
mxr: I see. Do the other six fixers handle this?

asottile: yes, this case differs because you have to inspect the arguments

pyupgrade.py
mxr: I was able to do it

PR #117, opened 2019-03-18
Remove six.Iterator (py3+) and combine six fix with class fix

mxr:
Part of #59.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -914,13 +914,11 @@ SIX_UNICODE_COMPATIBLE = 'python_2_unicode_compatible'
 
 
 class FindSixAndClassesUsage(ast.NodeVisitor):
-    Base = collections.namedtuple('Base', ('node', 'index'))
-
     def __init__(self):
         super(FindSixAndClassesUsage, self).__init__()
         self.call_attrs = {}
         self.call_names = {}
-        self.classes = {}
+        self.classes = set()
         self.raise_attrs = {}
         self.raise_names = {}
         self.simple_attrs = {}
@@ -956,20 +954,20 @@ class FindSixAndClassesUsage(ast.NodeVisitor):
 
         for i, base in enumerate(node.bases):
             if isinstance(base, ast.Name) and base.id == 'object':
-                self.classes[_ast_to_offset(base)] = self.Base(node, i)
+                self.classes.add(_ast_to_offset(base))
             elif (
                     isinstance(base, ast.Name) and
                     base.id == 'Iterator' and
                     base.id in self.six_from_imports
             ):
-                self.classes[_ast_to_offset(base)] = self.Base(node, i)
+                self.classes.add(_ast_to_offset(base))
             elif (
                     isinstance(base, ast.Attribute) and
                     isinstance(base.value, ast.Name) and
                     base.value.id == 'six' and
                     base.attr == 'Iterator'
             ):
-                self.classes[_ast_to_offset(base)] = self.Base(node, i)
+                self.classes.add(_ast_to_offset(base))
 
         self.generic_visit(node)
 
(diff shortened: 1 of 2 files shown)


tests/pyupgrade_test.py
```diff
@@ -1092,6 +1092,10 @@ def test_fix_six(s, expected):
             'class C(object, metaclass=ABCMeta): pass',
             'class C(metaclass=ABCMeta): pass',
         ),
+        (
+            'class C(six.Iterator, metaclass=ABCMeta): pass',
```
asottile: should also handle `from six import Iterator`  `class C(Iterator, ...):`

I wonder what happens if

```python
class C(Iterator, object):
```

🤔

mxr: Good point, let me try to handle those cases

hugovk:
@asottile I suggest updating the CI config to allow testing of all feature branches, so we know PRs pass tests.

asottile:
> @asottile I suggest updating the CI config to allow testing of all feature branches, so we know PRs pass tests.

weird, I wonder if azure just dropped the webhook here, it *is* supposed to test PRs (in fact, *only* PRs :laughing:): 

For instance https://github.com/asottile/pyupgrade/pull/115

though maybe that's because I ran it :thinking:

asottile:
ok I had to click a button, seems to work now: #118

hugovk:
I think removing the branch restriction would let it build all branches automatically:

https://github.com/asottile/pyupgrade/blob/0df1eac05c9c37e2bfecf11895aee53a34c33afd/azure-pipelines.yml#L2-L3

mxr:
@asottile let me know if the changes look good. I can't re-request a review on GitHub.

asottile:
github actually just added that as a feature:

Screenshot from 2019-03-20 08-25-39

(the little refresh button)

pyupgrade.py
```diff
@@ -852,26 +852,56 @@ def _fix_super(contents_text):
     return tokens_to_src(tokens)
 
 
-class FindNewStyleClasses(ast.NodeVisitor):
+class SixVisitor(ast.NodeVisitor):
```
asottile: instead of inheritance, let's combine `SixVisitor` and `FindsClasses` (I was actually going to do that as a followup anyway).  This will also improve performance of pyupgrade by ~15% as well :tada:

pyupgrade.py
```diff
@@ -852,26 +852,56 @@ def _fix_super(contents_text):
...
-            if isinstance(base, ast.Name) and base.id == 'object':
+            if (
+                (
+                    isinstance(base, ast.Name) and
+                    (
+                        base.id == 'object' or
+                        (
+                            base.id == 'Iterator' and
+                            'Iterator' in self.six_from_imports
+                        )
```
asottile: let's keep these cases separate just so it's a little easier to see what's going on -- I realize you golfed down the `isinstance(..., ast.Name)` but I think it's less readable now.

Also let's do `base.id in self.six_from_imports` for the last condition

pyupgrade.py
mxr: I'll have you know this is enterprise-grade inheritance

mxr:
I don't see  the little refresh button, only the "speech bubble" button :/

pyupgrade.py
mxr: What about FindSixUsage?

asottile:
> I don't see the little refresh button, only the "speech bubble" button :/

riiip, maybe you're not in the experiment bucket

pyupgrade.py
asottile: yeah, combine all 3

mxr:
@asottile combined all 3, lmk if that's what you had in mind

mxr:
@asottile do you have an expected merge timeline?

asottile:
> @asottile do you have an expected merge timeline?

I can't if the tests are failing :P

mxr:
confused the green approval check with the green tests passing check 😭 and the tests (that i ran) passed on my machine

asottile:
> confused the green approval check with the green tests passing check sob and the tests (that i ran) passed on my machine

looks like only python2.x is failing

pyupgrade.py
```diff
@@ -1034,13 +942,23 @@ def visit_ClassDef(self, node):
             ):
                 self.remove_decorators.add(_ast_to_offset(decorator))
 
-        self.generic_visit(node)
+        for i, base in enumerate(node.bases):  # pragma: no cover (py3+)
```
asottile: hmmmm, this seems not right -- it didn't need this comment on the left, why the change now?

mxr: Let me scope this down

mxr:
@asottile updated

mxr:
did i forget to push?

asottile:
> did i forget to push?

looks like it, yeah -- I still see the `no cover`s

mxr:
pushed 🤦‍♂️

asottile:
wompwomp

```python
import six


class C(object, six.Iterator): pass
```

```console
$ pyupgrade  t.py  --py3-plus
  File "/home/asottile/workspace/pyupgrade/venv/bin/pyupgrade", line 11, in <module>
    load_entry_point('pyupgrade', 'console_scripts', 'pyupgrade')()
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1402, in main
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1376, in fix_file
    contents_text = _fix_six_and_classes(contents_text)
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1135, in _fix_six_and_classes
    while tokens[j].src != ',':
IndexError: list index out of range
```

mxr:
something something unpaid labor

mxr:
@asottile still haven't had a chance to look at it. Would separating the six fixer and the class fixer be an option..?

asottile:
I don't think that helps -- I think all that needs to happen here is to make the class killer not worry about the class count and always look forward for parens or commas and then make a decision

mxr:
I see

mxr:
Thanks!!!

PR #149, opened 2019-05-22
Rewrite six.add_metaclass in --py3-plus mode

asottile:


mxr:
https://gfycat.com/shrillanimatedeasternnewt

PR #151, opened 2019-05-25
Implement 'yield from collection' formatter

### LINKED ISSUE #102: for x in y: yield x => yield from y
No need for `async for` support according to [this](https://www.python.org/dev/peps/pep-0525/#asynchronous-yield-from)

Here's an ast parser to detect this, just need the tokenization to rewrite it:

```python
class Visitor(ast.NodeVisitor):
    def __init__(self, filename):
        self.filename = filename

    def visit_For(self, node: ast.For):
        if (
            isinstance(node.target, ast.Name) and
            len(node.body) == 1 and
            isinstance(node.body[0], ast.Expr) and
            isinstance(node.body[0].value, ast.Yield) and
            isinstance(node.body[0].value.value, ast.Name) and
            node.target.id == node.body[0].value.value.id
        ):
            print(f'{self.filename}:{node.lineno}: could be yield from')

        self.generic_visit(node)
```

DanielChabrowski:
Resolves #102 

Known issues (?):
* Removes newlines from between for and yield
* triple-quoted strings prevent the upgrade

Another way to handle loop body would be to take everything, replace the yield, adjust indent. I think it would simplify code a little. Tell me what do you think :)

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -270,10 +270,20 @@ def _victims(tokens, start, arg, gen):
     return Victims(starts, ends, first_comma_index, arg_index)
 
 
-def _find_open_paren(tokens, i):
-    while tokens[i].src != '(':
+def _find_token(tokens, i, token):
+    while tokens[i].src != token:
         i += 1
     return i
+
+
+def _find_token_by_name(tokens, i, name):
+    while tokens[i].name != name:
+        i += 1
+    return i
+
+
+def _find_open_paren(tokens, i):
+    return _find_token(tokens, i, '(')
 
 
 def _is_on_a_line_by_self(tokens, i):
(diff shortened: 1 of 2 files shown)


asottile:
🎉 good job on this!

here's what it does to pytest for instance:

```diff
diff --git a/src/_pytest/main.py b/src/_pytest/main.py
index 5b00e65c3..3854dd71a 100644
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -515,8 +515,7 @@ class Session(nodes.FSCollector):
             self.trace("processing argument", arg)
             self.trace.root.indent += 1
             try:
-                for x in self._collect(arg):
-                    yield x
+                yield from self._collect(arg)
             except NoMatch:
                 # we are inside a make_report hook so
                 # we cannot directly pass through the exception
@@ -598,8 +597,7 @@ class Session(nodes.FSCollector):
             if argpath.basename == "__init__.py":
                 yield next(m[0].collect())
                 return
-            for y in m:
-                yield y
+            yield from m
 
     def _collectfile(self, path, handle_dupes=True):
         assert path.isfile(), "{!r} is not a file (isdir={!r}, exists={!r}, islink={!r})".format(
@@ -737,6 +735,5 @@ class Session(nodes.FSCollector):
             rep = collect_one_node(node)
             if rep.passed:
                 for subnode in rep.result:
-                    for x in self.genitems(subnode):
-                        yield x
+                    yield from self.genitems(subnode)
             node.ihook.pytest_collectreport(report=rep)
diff --git a/src/_pytest/python.py b/src/_pytest/python.py
index 01b1b8c02..a1213e1be 100644
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -656,8 +656,7 @@ class Package(Module):
                 continue
 
             if is_file:
-                for x in self._collectfile(path):
-                    yield x
+                yield from self._collectfile(path)
             elif not path.isdir():
                 # Broken symlink or invalid/missing file.
                 continue
```

Issue #138, opened 2019-04-30
py3-plus: .encode('utf-8') → .encode() & .decode('utf-8') → .decode()


jdufresne:
In Python 3, the first positional argument to `.encode()` and `.decode()` defaults to `'utf-8'`. This was not the case with Python 2:

https://docs.python.org/3/library/stdtypes.html#str.encode
https://docs.python.org/3/library/stdtypes.html#bytes.decode

As this is the most common value passed, there could be a transform to simplify the calls to just `.encode()` or `.decode()` (without the first positional argument). This could help clean up and simplify modern Python code by removing noise.

asottile:
hmm github must have lost my comment I had typed something out but it's gone

I'm not sure this rewrite is safe in all cases, since an arbitrary object could implement `decode` / `encode` to have different semantics

for instance `webob.Request` has a `decode` method

mxr:
Is it worth doing when the receiver is a string literal?

asottile:
I think the only case that could safely be rewritten is for literals and `.encode()`

```python
"unicode text☃".encode('UTF-8')   # "unicode text☃".encode()
"ascii text".encode('UTF-8')             # b"ascii text"
```

mxr:
I see. Why can't `b'asd'.decode('UTF-8')` be safely transformed? Can pyupgrade try to execute the decoding & only apply the transformation if decoding doesn't raise?

asottile:
it can, but like, who would write that

WSLUser:
Will this work if pyupgrade has already been used (along with modernize and/or futurize)?

asottile:
> Will this work if pyupgrade has already been used (along with modernize and/or futurize)?

Could you elaborate? `pyupgrade` is idempotent and doesn't have an opinion on other tools that are used

WSLUser:
I mean if I ran pyupgrade for a script already, will it run through and successfully make the changes using this new addition for the same script?

asottile:
yes, however this particular feature has not been released yet

WSLUser:
Ah, I saw the docs were updated and assumed it was available as a beta at least on PyPi.

asottile:
[...]

PR #154, opened 2019-05-27
Rewrite `.encode('utf-8')` in some cases

mxr:
Attempts to address #138

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import argparse
 import ast
+import codecs
 import collections
 import io
 import keyword
@@ -942,6 +943,13 @@ SIX_RAISES = {
 }
 
 
+def _is_utf8_codec(encoding):
+    try:
+        return codecs.lookup(encoding).name == 'utf-8'
+    except LookupError:
+        return False
+
+
 class FindPy3Plus(ast.NodeVisitor):
     class ClassInfo:
         def __init__(self, name):
@@ -952,6 +960,7 @@ class FindPy3Plus(ast.NodeVisitor):
     def __init__(self):
         self.bases_to_remove = set()
 
+        self.encode_calls = {}
         self.native_literals = set()
 
         self._six_from_imports = set()
@@ -1084,6 +1093,16 @@ class FindPy3Plus(ast.NodeVisitor):
                 not _starargs(node)
         ):
             self.native_literals.add(_ast_to_offset(node))
+        elif (
+                isinstance(node.func, ast.Attribute) and
+                isinstance(node.func.value, ast.Str) and
+                node.func.attr == 'encode' and
+                not _starargs(node) and
+                len(node.args) == 1 and
+                isinstance(node.args[0], ast.Str) and
+                _is_utf8_codec(node.args[0].s)
+        ):
+            self.encode_calls[_ast_to_offset(node)] = node
 
         self.generic_visit(node)
 
@@ -1214,6 +1233,7 @@ def _fix_py3_plus(contents_text):
 
     if not any((
             visitor.bases_to_remove,
+            visitor.encode_calls,
             visitor.native_literals,
             visitor.six_add_metaclass,
             visitor.six_b,
@@ -1319,6 +1339,11 @@ def _fix_py3_plus(contents_text):
         elif token.offset in visitor.super_calls:
             i = _find_open_paren(tokens, i)
             call = visitor.super_calls[token.offset]
+            victims = _victims(tokens, i, call, gen=False)
+            del tokens[victims.starts[0] + 1:victims.ends[-1]]
+        elif token.offset in visitor.encode_calls:
+            i = _find_open_paren(tokens, i)
+            call = visitor.encode_calls[token.offset]
             victims = _victims(tokens, i, call, gen=False)
             del tokens[victims.starts[0] + 1:victims.ends[-1]]
         elif token.offset in visitor.native_literals:
--- a/tests/pyupgrade_test.py
+++ b/tests/pyupgrade_test.py
@@ -1449,6 +1449,41 @@ def test_fix_native_literals(s, expected):
 
 
 @pytest.mark.parametrize(
+    ('s', 'expected'),
+    (
+        ('"asd".encode("utf-8")', '"asd".encode()'),
+        ('"asd".encode("utf8")', '"asd".encode()'),
+        ('"asd".encode("UTF-8")', '"asd".encode()'),
+        (
+            'sys.stdout.buffer.write(\n    "a"\n    "b".encode("utf-8")\n)',
+            'sys.stdout.buffer.write(\n    "a"\n    "b".encode()\n)',
+        ),
+    ),
+)
+def test_fix_encode(s, expected):
+    assert _fix_py3_plus(s) == expected
+
+
+@pytest.mark.parametrize(
+    's',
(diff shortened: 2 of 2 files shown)


pyupgrade.py
```diff
@@ -1084,6 +1085,27 @@ def visit_Call(self, node):  # type: (ast.Call) -> None
...
         ):
             self.native_literals.add(_ast_to_offset(node))
+        elif (
+                isinstance(node.func, ast.Attribute) and
+                isinstance(node.func.value, ast.Str) and
+                node.func.attr == 'encode' and
+                not _starargs(node) and
+                len(node.args) == 1 and
+                isinstance(node.args[0], ast.Str) and
+                node.args[0].s.lower() == 'utf-8'
```
asottile: `'utf8'` works too -- there's a function in either `codecs` or `encodings` that normalizes this value -- can probably reuse that here?

pyupgrade.py
```diff
@@ -1084,6 +1085,27 @@ def visit_Call(self, node):  # type: (ast.Call) -> None
...
+        elif (
+                isinstance(node.func, ast.Attribute) and
+                isinstance(node.func.value, ast.Str) and
+                node.func.attr == 'encode' and
+                not _starargs(node) and
+                len(node.keywords) == 1 and
+                node.keywords[0].arg == 'encoding' and
+                isinstance(node.keywords[0].value, ast.Str) and
+                node.keywords[0].value.s.lower() == 'utf-8'
+        ):
```
asottile: I wouldn't bother with this `if`, I don't think I've ever seen `.encode(encoding='UTF-8')`

pyupgrade.py
```diff
@@ -1224,6 +1246,7 @@ def _fix_py3_plus(contents_text):
             visitor.six_type_ctx,
             visitor.six_with_metaclass,
             visitor.super_calls,
+            visitor.encode_calls,
```
asottile: these are alphabetized, as are the attributes on `Py3PlusVisitor` (~roughly -- though grouped with their relevant private state)

tests/pyupgrade_test.py
```diff
@@ -1448,6 +1448,30 @@ def test_fix_native_literals(s, expected):
...
+        ('"asd".encode(encoding="utf-8")', 'b"asd"'),
+        ('"asd".encode(encoding="UTF-8")', 'b"asd"'),
+
+        ('"☕️".encode("utf-8")', '"☕️".encode()'),
+        ('"☕️".encode("UTF-8")', '"☕️".encode()'),
+        ('"☕️".encode(encoding="utf-8")', '"☕️".encode()'),
+        ('"☕️".encode(encoding="UTF-8")', '"☕️".encode()'),
+
+        # the current version only works for string literals
+        ('x="asd"\nx.encode("utf-8")', 'x="asd"\nx.encode("utf-8")'),
```
asottile: the general trend I've done is to add a separate parametrized `_noop` test so it's easier to see

tests/pyupgrade_test.py
```diff
@@ -1448,6 +1448,30 @@ def test_fix_native_literals(s, expected):
     assert _fix_py3_plus(s) == expected
 
 
+@pytest.mark.parametrize(
+    ('s', 'expected'),
+    (
+        ('"asd".encode("utf-8")', 'b"asd"'),
```
asottile: curious what does this does for:

```python
'a' 'b'.encode('UTF-8')

sys.stdout.buffer.write(
    'a'
    'b'.encode('UTF-8')
)
```

pyupgrade.py
mxr: OK

pyupgrade.py
mxr: Sure

mxr: Hmm I found this one but it seems less than useful

```
In [5]: encodings.normalize_encoding('utf-8')
Out[5]: 'utf_8'

In [6]: encodings.normalize_encoding('utf8')
Out[6]: 'utf8'

In [7]: encodings.normalize_encoding('UTF8')
Out[7]: 'UTF8'
```

I wasn't able to find anything in codecs (from ipython autocomplete). Any advice?

pyupgrade.py
mxr: Sure

tests/pyupgrade_test.py
mxr: sure

tests/pyupgrade_test.py
mxr: Looks like my change causes mixing of byte & nonbyte literals. I could not handle this case at all or I could modify the predecessor to be a byte string if possible?

pyupgrade.py
asottile:
```pycon
>>> codecs.lookup('utf_8').name
'utf-8'
>>> codecs.lookup('utf8').name
'utf-8'
>>> codecs.lookup('UTF8').name
'utf-8'
```

tests/pyupgrade_test.py
asottile: maybe let's not convert to bytestrings for now then, and just remove the encoding name -- I think that makes your change simpler / safer anyway 👍

mxr: suree

pyupgrade.py
mxr: ty!

tests/pyupgrade_test.py
mxr: sure* `!sux mbp`

mxr:
Can't request a review using GitHub for some reason but this is ready for another 👀

pyupgrade.py
```diff
@@ -1092,6 +1104,13 @@ def generic_visit(self, node):
         super(FindPy3Plus, self).generic_visit(node)
 
 
+def _is_utf8_codec(encoding):
+    try:
+        return codecs.lookup(encoding).name == 'utf-8'
+    except LookupError:
+        return False
```
asottile: I tend to put helpers _higher_ than their callers in the file, nbd though

tests/pyupgrade_test.py
```diff
@@ -1448,6 +1448,39 @@ def test_fix_native_literals(s, expected):
...
+    ),
+)
+def test_fix_encode(s, expected):
+    assert _fix_py3_plus(s) == expected
+
+
+@pytest.mark.parametrize(
+    's',
+    (
+        # see https://git.io/fjRcs
```
asottile: let's just write out the comment, the link doesn't seem to work anyway
```python
# for now, pyupgrade is timid in this particular case
```

asottile: could also just move this down below the other comment about timidity (is that a word?)

mxr: sure

asottile:
also looks like you need a test case for

```python
'foo'.encode('unknown-encoding')
```

pyupgrade.py

mxr:
Yea I didn't run coverage locally :( added one in the latest commit

pyupgrade.py
```diff
@@ -981,6 +983,12 @@ def _is_six(self, node, names):
             node.attr in names
         )
 
+    def _is_utf8_codec(self, encoding):
```
asottile: eheh not exactly what I meant -- since it doesn't make sense as a method (doesn't use `self`)

mxr: I can move it

PR #168, opened 2019-06-15
Fix yield from for async functions

### LINKED ISSUE #167: Should not rewrite to `yield from` inside `async def`
oops, totally forgot about this.

Currently this is happening but this is forbidden by [PEP 525](https://www.python.org/dev/peps/pep-0525/#asynchronous-yield-from)

```diff
 async def f():
-    for x in y:
-        yield x
+    yield from y
```

DanielChabrowski:
Fix #167

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1109,8 +1109,7 @@ class FindPy3Plus(ast.NodeVisitor):
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
 
-    # pragma: no cover (py35+)
-    def visit_AsyncFunctionDef(self, node):
+    def visit_AsyncFunctionDef(self, node):  # pragma: no cover (py35+)
         self._in_async_def, orig = True, self._in_async_def
         self._visit_func(node)
         self._in_async_def = orig
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1022,6 +1022,7 @@ class FindPy3Plus(ast.NodeVisitor):
 
         self._class_info_stack = []
         self._in_comp = 0
+        self._in_async_def = False
         self.super_calls = {}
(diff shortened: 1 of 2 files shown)


pyupgrade.py
```diff
@@ -1022,6 +1022,7 @@ def __init__(self):
 
         self._class_info_stack = []
         self._in_comp = 0
+        self._in_async_def = False
```
asottile: move this down one line so that the state is associated with the thing it needs (for instance `_in_comp` is affecting `super_calls`)

pyupgrade.py
```diff
@@ -1101,7 +1102,17 @@ def _visit_func(self, node):
...
 
-    visit_FunctionDef = visit_Lambda = _visit_func
+    def _visit_sync_func(self, node):
+        self._in_async_def, orig = False, self._in_async_def
+        self._visit_func(node)
+        self._in_async_def = orig
+
+    visit_FunctionDef = visit_Lambda = _visit_sync_func
+
+    def visit_AsyncFunctionDef(self, node):
```
asottile: you'll need to `# pragma: no cover PY35+` this function since it's version dependent

tests/pyupgrade_test.py
```diff
@@ -1572,6 +1610,27 @@ def test_fix_yield_from_noop(s):
...
+        '        yield x',
+        'async def f():\n'
+        '    def g():\n'
+        '        async def z():\n'
+        '            for x in [1, 2, 3]:\n'
+        '                yield x',
+    ),
+)
+def test_fix_async_yield_from_noop(s):
+    assert _fix_py3_plus(s) == s
```
asottile: I think you can delete this test, it's covered above

Issue #141, opened 2019-05-10
py3.3+: IOError (and others) -> OSError


jdufresne:
In Python 3.3+ the following exceptions are now simple aliases of `OSError`: `IOError`, `EnvironmentError`, `WindowsError`, `mmap.error`, `socket.error` and `select.error`.

The aliases have been removed from the documented hierarchy: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

I think pyupgrade should rewrite these aliases as the canonical `OSError`.

References:

https://docs.python.org/3/whatsnew/3.3.html#pep-3151-reworking-the-os-and-io-exception-hierarchy

https://docs.python.org/3/library/exceptions.html#OSError

asottile:
@mxr did some work on this (that just got released) (replacing just the plain aliases for now)

Another thing that would be nice to have (noticed while trying this on `pytest`):

(special case the `except` block type tuple)

```python
try:
    ...
except (IOError, OSError):    # except OSError:
    ...
```

mxr:
I was hacking on that a bit today but there are a ton of fun cases

```
except (IOError, OSError):				# except OSError:
except(IOError, OSError):				# except OSError:
except(IOError,):						# except OSError:
except (IOError, KeyError, OSError):	# except (OSError, KeyError): (how do we even choose the right order)
```

etc

asottile:
those all seem correct to me, I don't think ordering is important for the last one

mxr:
Anything else left for this ticket?

asottile:
I don't think so, I might make a second pass where `pyupgrade` properly knows what things are global or reassinged or what-not so we can cover calls as well as just the `except` / `raise` blocks that we have higher confidence about

PR #170, opened 2019-06-21
Rewrite some OSError aliases

mxr:
Attempts to address part of #141

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -974,6 +974,12 @@ def _is_utf8_codec(encoding):
 
 
 class FindPy3Plus(ast.NodeVisitor):
+    OS_ERROR_ALIASES = frozenset((
+        'EnvironmentError',
+        'IOError',
+        'WindowsError',
+    ))
+
     class ClassInfo:
         def __init__(self, name):
             self.name = name
@@ -990,6 +996,7 @@ class FindPy3Plus(ast.NodeVisitor):
         self.if_py3_blocks = set()
 
         self.native_literals = set()
+        self.simple_ids = {}
 
         self._six_from_imports = set()
         self.six_add_metaclass = set()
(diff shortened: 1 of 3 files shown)


asottile:
we don't support py32 so I think you can just use `--py3-plus`

mxr:
OK will change it. Also do you think I should use `visit_Raise` to only rewrite usages in `raise` statements?

pyupgrade.py
```diff
@@ -1785,6 +1785,54 @@ def _fix_fstrings(contents_text):
     return tokens_to_src(tokens)
 
 
+class FindOSErrorAliases(ast.NodeVisitor):
```
asottile: let's combine this with the Py3Plus visitor -- each visitor we add adds ~30% execution time, but if we batch them up it's ~1-3% instead

mxr: Sure

asottile:
hmm yeah I wonder if it should only be in `raise` / `except` :thinking:  -- I guess the only downside would be potentially rewriting unrelated things

mxr:
It does feel like a broad rewrite but I'm not sure if I can think of a specific example that's a needless rewrite with this implementation.

Also apologies to Satya for hammering CI with my busted code

pyupgrade.py
mxr: What's the slow part btw? Finding the nodes or parsing from src to tokens and back? I wonder if there's a way to parallelize the work in a way that you could write a single "finder/replacer" each thing to rewrite, instead of having one big one, so you could keep the implementation fast but still manageable

asottile: most of it comes from ast parse, the tokenize-untokenize is not the fastest as well

mxr:
Think this is ready for a re-review (still using visit_Name)

mxr:
Provided `ModuleNotFoundError: No module named '_sqlite3'` is not from this change

mxr:
Any thoughts on the sqlite error?

asottile:
yeah, azure pipelines broke python2.7, 3.5, 3.7 and they still haven't completely rolled it back -- safe to ignore

mxr:
OK. I think everything else is green unless I misread the output

PR #171, opened 2019-06-21
Add type annotations

mxr:


pyupgrade.py
```diff
@@ -22,26 +22,77 @@
...
 
+try:
+    from typing import Text
+except ImportError:  # pragma: no cover
+    if sys.version_info < (3, 0):
+        Text = unicode  # type: ignore  # noqa
+    else:
+        Text = str  # type: ignore
+
+if False:
```
mxr: Should typing be a new dep under py2?

.pre-commit-config.yaml
```diff
@@ -29,3 +29,11 @@ repos:
     rev: v1.19.0
     hooks:
     -   id: pyupgrade
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.701
+    hooks:
+    -   id: mypy
+    -   id: mypy
+        args:
+        -   --py2
```
mxr: A lot of the ignores/hacks were to support py2. Do we care about running mypy under py2?

setup.cfg
```diff
@@ -30,5 +30,22 @@ python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
 console_scripts =
     pyupgrade = pyupgrade:main
 
+[options.package_data]
+pyupgrade =
+    py.typed
```
mxr: Not sure if anyone is importing pyupgrade but it doesn't hurt to provide the stubs

setup.cfg
```diff
@@ -30,5 +30,22 @@ python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
...
+
 [bdist_wheel]
 universal = True
+
+[mypy]
```
mxr: This config was copied from all-repos

pyupgrade.py
asottile: yes

pyupgrade.py
```diff
@@ -458,6 +524,8 @@ def _imports_unicode_literals(contents_text):
         else:
             return False
 
+    return False
```
mxr: mypy caught this

setup.cfg
asottile: we can't we're a single file and not a package

pyupgrade.py
asottile: not sure this is possible 🤷‍♂

.pre-commit-config.yaml
asottile: no, we do not

tests/pyupgrade_test.py
```diff
@@ -47,7 +48,7 @@ def test_roundtrip_text(s):
 )
 def test_intentionally_not_round_trip(s, expected):
     # Our unparse simplifies empty parts, whereas stdlib allows them
-    ret = unparse_parsed_string(parse_format(s))
+    ret = unparse_parsed_string(list(parse_format(s)))
```
asottile: `list()` shouldn't be necessary here

tests/pyupgrade_test.py
```diff
@@ -753,9 +754,13 @@ def test_percent_format_noop(s):
     assert _fix_percent_format(s) == s
 
 
+def _get_body(expr):
+    return cast(ast.Expr, cast(ast.Module, ast.parse(expr)).body[0]).value
```
asottile: gross :/

pyupgrade.py
```diff
@@ -22,26 +22,77 @@
 from tokenize_rt import tokens_to_src
 from tokenize_rt import UNIMPORTANT_WS
 
+try:
+    from typing import Text
+except ImportError:  # pragma: no cover
+    if sys.version_info < (3, 0):
+        Text = unicode  # type: ignore  # noqa
+    else:
+        Text = str  # type: ignore
```
asottile: no need for this, you import `Text` below

pyupgrade.py
```diff
@@ -13,7 +13,7 @@
 import tokenize
 import warnings
 
-from tokenize_rt import NON_CODING_TOKENS
+from tokenize_rt import NON_CODING_TOKENS  # type: ignore
```
asottile: why is this type ignored?

pyupgrade.py
```diff
@@ -22,26 +22,77 @@
...
-def parse_format(s):
+def parse_format(s):  # type: (Text) -> Tuple[Parsed, ...]
     """Makes the empty string not a special case.  In the stdlib, there's
     loss of information (the type) on the empty string.
     """
-    parsed = tuple(_stdlib_parse_format(s))
+    parsed = tuple(_stdlib_parse_format(s))  # type: ignore
     if not parsed:
-        return ((s, None, None, None),)
+        return (s, None, None, None),
```
asottile: the parens are more readable here

pyupgrade.py
```diff
@@ -22,26 +22,77 @@
...
     else:
         return parsed
 
 
-def unparse_parsed_string(parsed):
+def unparse_parsed_string(parsed):  # type: (List[Parsed]) -> Text
```
asottile: should be `Sequence[Parsed]`

pyupgrade.py
```diff
@@ -57,26 +108,29 @@ def _convert_tup(tup):
...
     return Offset(node.lineno, node.col_offset)
 
 
-def ast_parse(contents_text):
+def ast_parse(contents_text):  # type: (Text) -> ast.Module
     # intentionally ignore warnings, we might be fixing warning-ridden syntax
     with warnings.catch_warnings():
-        warnings.simplefilter('ignore')
-        return ast.parse(contents_text.encode('UTF-8'))
+        warnings.simplefilter('ignore')  # type: ignore
```
asottile: weird, do the stubs really not have `simplefilter`? can you link a bug for each `# type: ignore`

pyupgrade.py
```diff
@@ -57,26 +108,29 @@ def _convert_tup(tup):
...
+        ast_module = ast.parse(contents_text.encode('UTF-8'))
 
+    assert isinstance(ast_module, ast.Module)
```
asottile: weird, when is this ever false?

pyupgrade.py
```diff
@@ -57,26 +108,29 @@ def _convert_tup(tup):
...
+    return ast_module
 
-def inty(s):
+
+def inty(s):  # type: (Any) -> bool
```
asottile: should be `# type: (str) -> bool`

pyupgrade.py
```diff
@@ -349,17 +415,17 @@ def _process_is_literal(tokens, i, compare):
...
     LITERAL_TYPES += (ast.Bytes,)
 
 
 class Py2CompatibleVisitor(ast.NodeVisitor):
     def __init__(self):  # type: () -> None
-        self.dicts = {}
-        self.sets = {}
-        self.set_empty_literals = {}
-        self.is_literal = {}
+        self.dicts = {}  # type: Dict[Offset, Union[ast.GeneratorExp, ast.ListComp]]  # noqa
```
asottile: what's the `noqa` here?

pyupgrade.py
```diff
@@ -385,7 +451,7 @@ def visit_Call(self, node):  # type: (ast.Call) -> None
                 len(node.args[0].elt.elts) == 2
         ):
             arg, = node.args
-            self.dicts[_ast_to_offset(node.func)] = arg
+            self.dicts[_ast_to_offset(node.func)] = arg  # type: ignore
```
asottile: what's the ignore here?

mxr:
Thanks for the feedback. I'll try to get back to this before Monday and give more descriptive answers about the ignores/noqas in the next iteration

setup.cfg
mxr: :mindblown:

pyupgrade.py
mxr: it was a str/unicode issue (py2 mypy)

pyupgrade.py
mxr: 1. line length

pyupgrade.py
mxr: i tried to keep it generic

pyupgrade.py
mxr: arg is inferred as an ast.expr

pyupgrade.py
mxr: py2 mypy couldn't find `tokenize_rt`

pyupgrade.py
mxr: i figured the type annotation would add clarity instead but i can put this back

pyupgrade.py
mxr: since it gracefully handles non-strs. also note the caller could pass null originally.

.pre-commit-config.yaml
mxr: wonderful

pyupgrade.py
mxr: tell it to mypy? :/

tests/pyupgrade_test.py
mxr: `unparse_parsed_string`'s argument takes a list but i'll try to rework this

pyupgrade.py
```diff
@@ -1827,14 +1931,14 @@ def fix_file(filename, args):
 
     if contents_text != contents_text_orig:
         print('Rewriting {}'.format(filename))
-        with io.open(filename, 'w', encoding='UTF-8', newline='') as f:
-            f.write(contents_text)
+        with io.open(filename, 'w', encoding='UTF-8', newline='') as f:  # type: ignore  # noqa
```
mxr: OK this was super nasty. Had to pop into the mypy debugger to figure out what was going on. Since we use an `open()` call above with `rb`, that `f` is resolved to a BinaryIO, and here the same named variable `f` also gets resolved to `BinaryIO`. Will file a ticket upstream.

mxr: The fix would be to use `with open(...) as fb` above or use `io.open` everywhere

asottile: ah yeah can just use a different variable name, this is the usual "redefinition with different type" issue

pyupgrade.py
mxr: always. I could use a cast instead?

mxr: I mean it's always true

asottile: yeah I just wonder why typeshed doesn't have `-> Module` for `ast.parse`

pyupgrade.py
mxr: I think there is a new/specific interaction between open and io.open especially given that ond fix is to not mix them.  f is a new variable each time with its own scope right?

asottile: nope, variables are function scoped in python (except for `except`'s `as` in python3+ -- and even that is only simulated through `del` shenanigans)

tests/pyupgrade_test.py
mxr: https://github.com/python/typeshed/pull/3039

pyupgrade.py
mxr: :(

mxr:
@asottile I think this is ready for another review. Try not to embarrass me too much on stream though

pyupgrade.py
mxr: Left a link to a typeshed PR that fixes this IIUC

.coveragerc
```diff
@@ -27,6 +27,9 @@ exclude_lines =
     # Don't complain if non-runnable code isn't run:
     ^if __name__ == ['"]__main__['"]:$
 
+    # Don't complain about type hints
+    ^if TYPE_CHECKING:$
```
asottile: you'll need to use `if False` since python3.5 is still within our support (for now) -- fortunately `coverage` knows about `if False` by default so you won't need to edit this file after all!

pyupgrade.py
```diff
@@ -12,6 +12,23 @@
...
+from typing import Dict
+from typing import Iterable
+from typing import Iterator
+from typing import List
+from typing import Match
+from typing import Optional
+from typing import Pattern
+from typing import Sequence
+from typing import Set
+from typing import Text
```
asottile: can just use `str` since I don't care about python2 mypy

pyupgrade.py
```diff
@@ -23,10 +40,29 @@
 from tokenize_rt import UNIMPORTANT_WS
 
 
+Parsed = Tuple[Text, Optional[Text], Optional[Text], Optional[Text]]
+Fmt = Tuple[
+    Optional[Text],
+    Optional[Text],
+    Optional[Text],
+    Optional[Text],
+    Optional[Text],
```
asottile: I think this one is non-optional

pyupgrade.py
```diff
@@ -23,10 +40,29 @@
 from tokenize_rt import UNIMPORTANT_WS
 
 
+Parsed = Tuple[Text, Optional[Text], Optional[Text], Optional[Text]]
+Fmt = Tuple[
```
asottile: maybe `DotFormatPart` and `PercentFormatPart`?

pyupgrade.py
```diff
@@ -23,10 +40,29 @@
...
+    if sys.version_info >= (3,):
+        AsyncFunctionDef = ast.AsyncFunctionDef
+    else:
+        AsyncFunctionDef = ast.stmt
+
 _stdlib_parse_format = string.Formatter().parse
 
 
-def parse_format(s):
+def parse_format(s):  # type: (Text) -> Tuple[Parsed, ...]
```
asottile: lol, not worth fixing in this branch -- but I went through all the work to preserve the types here and it's not even possible that this receives bytes in python2 now that we're using `tokenize-rt`

pyupgrade.py
asottile: hmmm, how about `Optional[str] -> bool` then?

pyupgrade.py
```diff
@@ -335,6 +382,7 @@ def _process_dict_comp(tokens, start, arg):
 
 
 def _process_is_literal(tokens, i, compare):
+    # type: (List[Token], int, ast.cmpop) -> None
```
asottile: should this be `Union[ast.Is, ast.IsNot]`?

pyupgrade.py
asottile: oh it definitely is possible, heh -- the testsuite even exercises it!

pyupgrade.py
```diff
@@ -615,8 +664,17 @@ def _fix_tokens(contents_text, py3_plus):
...
+    # type: (Pattern[Text], Text, int) -> Match[Text]
+    match = regex.match(string, pos)
+    assert match is not None
+    return match
+
+
 def parse_percent_format(s):
+    # type: (Text) -> Tuple[Tuple[Text, Optional[Fmt]], ...]
     def _parse_inner():
+        # type: () -> Iterator[Tuple[Text, Optional[Fmt]]]
```
asottile: this one should be `Generator` as well

it is true that Generators are Iterators, but there are situations where an iterator cannot be used in place of a generator (for instance `@contextlib.contextmanager` or using `.send()`, etc.)

pyupgrade.py
```diff
@@ -791,9 +850,14 @@ def _fix_percent_format_tuple(tokens, start, node):
...
 
 def _fix_percent_format_dict(tokens, start, node):
-    seen_keys = set()
+    # type: (List[Token], int, ast.BinOp) -> None
+    seen_keys = set()  # type: Set[Text]
     keys = {}
-    for k in node.right.keys:
+
+    # the caller has enforced this, but we have to tell mypy again
+    right = cast(ast.Dict, node.right)
```
asottile: I usually prefer `assert` instead of `cast` -- it has the same affect but with the added benefit of being checked

pyupgrade.py
```diff
@@ -819,17 +883,17 @@ def _fix_percent_format_dict(tokens, start, node):
 
     key_indices = []
     for i, token in enumerate(tokens[brace:brace_end], brace):
-        k = keys.pop(token.offset, None)
-        if k is None:
+        key = keys.pop(token.offset, None)  # type: Optional[ast.Str]
```
asottile: I'm surprised this requires an annotation 🤔

pyupgrade.py
```diff
@@ -1143,14 +1215,14 @@ def visit_Call(self, node):  # type: (ast.Call) -> None
...
+                last_frame and
+                last_frame.def_depth == 1 and
                 isinstance(node.func, ast.Name) and
                 node.func.id == 'super' and
                 len(node.args) == 2 and
                 all(isinstance(arg, ast.Name) for arg in node.args) and
-                node.args[0].id == self._class_info_stack[-1].name and
-                node.args[1].id == self._class_info_stack[-1].first_arg_name
+                cast(ast.Name, node.args[0]).id == last_frame.name and
+                cast(ast.Name, node.args[1]).id == last_frame.first_arg_name
```
asottile: instead of `cast`, could conver the cheeky `all(...)` into two `isinstance` calls

pyupgrade.py
```diff
@@ -1124,13 +1193,16 @@ def visit_Name(self, node):  # type: (ast.Name) -> None
         self._visit_simple(node)
 
     def visit_Call(self, node):  # type: (ast.Call) -> None
+        last_frame = self._class_info_stack and self._class_info_stack[-1]
```
asottile: it's not clear to me why this was moved out, now we're computing this on every invocation instead of just when it matters

pyupgrade.py
```diff
@@ -1240,12 +1320,16 @@ def visit_If(self, node):  # type: (ast.If) -> None
...
-            isinstance(node.body[0].value, ast.Yield) and
-            targets_same(node.target, node.body[0].value.value) and
-            not node.orelse
+                not self._in_async_def and
+                len(node.body) == 1 and
+                isinstance(node.body[0], ast.Expr) and
+                isinstance(node.body[0].value, ast.Yield) and
+                targets_same(
+                    node.target,
+                    # node.body[0].value.value cannot be None here
```
asottile: it actually can be, interesting...

```console
$ pyupgrade t.py  --py3-plus
  File "/private/tmp/venv/bin/pyupgrade", line 10, in <module>
    sys.exit(main())
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1830, in main
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1804, in fix_file
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1503, in _fix_py3_plus
    visitor.visit(ast_obj)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 262, in visit
    return visitor(node)
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1244, in generic_visit
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 270, in generic_visit
    self.visit(item)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 262, in visit
    return visitor(node)
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1089, in _visit_sync_func
    self._visit_func(node)
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1085, in _visit_func
    self.generic_visit(node)
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1244, in generic_visit
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 270, in generic_visit
    self.visit(item)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 262, in visit
    return visitor(node)
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 1235, in visit_For
    targets_same(node.target, node.body[0].value.value) and
  File "/private/tmp/venv/lib/python3.7/site-packages/pyupgrade.py", line 957, in targets_same
    for t1, t2 in zip(ast.walk(target), ast.walk(yield_value)):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 234, in walk
    todo.extend(iter_child_nodes(node))
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 189, in iter_child_nodes
    for name, field in iter_fields(node):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ast.py", line 177, in iter_fields
    for field in node._fields:
AttributeError: 'NoneType' object has no attribute '_fields'
$ cat t.py 
def f():
    for x in y:
        yield
```

pyupgrade.py
```diff
@@ -1287,27 +1371,25 @@ class Block(
...
+    def _initial_indent(self, tokens):  # type: (List[Token]) -> int
         if tokens[self.start].src.isspace():
             return len(tokens[self.start].src)
         else:
             return 0
 
-    def _minimum_indent(self, tokens):
-        block_indent = None
+    def _minimum_indent(self, tokens):  # type: (List[Token]) -> int
+        block_indent = sys.maxsize
```
asottile: oof this is going to do some nasty things if this function is broken -- I'd rather keep this `None` and if you need to convince mypy then `assert block_indent is not None` at the end of the function

pyupgrade.py
```diff
@@ -1593,13 +1680,13 @@ def _replace(i, mapping, node):
             j = _find_open_paren(tokens, i)
             func_args, end = _parse_call_args(tokens, j)
             node = visitor.six_calls[token.offset]
-            template = _get_tmpl(SIX_CALLS, node.func)
+            template = _get_tmpl(SIX_CALLS, cast(ast.Attribute, node.func))
```
asottile: this isn't always an `Attribute` right? I thought it was `Union[Name, Attribute]`

pyupgrade.py
```diff
@@ -1689,18 +1776,19 @@ def _starargs(call):
...
 
-def _format_params(call):
+def _format_params(call):  # type: (ast.Call) -> Dict[Text, Text]
     params = {}
     for i, arg in enumerate(call.args):
-        params[str(i)] = _unparse(arg)
+        params[Text(i)] = _unparse(arg)
     for kwd in call.keywords:
-        params[kwd.arg] = _unparse(kwd.value)
+        # kwd.arg can't be None here
```
asottile: because we exclude starargs

pyupgrade.py
```diff
@@ -1805,8 +1893,8 @@ def _fix_fstrings(contents_text):
     return tokens_to_src(tokens)
 
 
-def fix_file(filename, args):
-    with open(filename, 'rb') as f:
+def fix_file(filename, args):  # type: (Text, argparse.Namespace) -> int
+    with io.open(filename, 'rb') as f:
```
asottile: hehe, I have another fixer I've been working on that auto-rewrites `io.open(..., 'rb')` => `open(..., 'rb')` so I'd rather rename one of the `f` variables

tests/pyupgrade_test.py
```diff
@@ -753,9 +754,13 @@ def test_percent_format_noop(s):
     assert _fix_percent_format(s) == s
 
 
+def _get_body(expr):
+    return cast(ast.Expr, ast.parse(expr).body[0]).value
```
asottile: I'd rather `assert isinstance(...)`

pyupgrade.py
asottile: also please add `flake8-typing-imports` so I can make sure to not break people on old py35

pyupgrade.py
mxr: It's a hack to support https://github.com/asottile/pyupgrade/blob/c0642054382847c707ec65ec070ffadd89c3dabd/pyupgrade.py#L736

pyupgrade.py
mxr: Not following, are you saying to create 2 types? And use them separately from Fmt?

pyupgrade.py
mxr: I can change it but I'm not sure I agree. Shouldn't typing be as loose as possible based on the function's internals, instead of on how the callers use it? For example suppose we have this function which is always called with a list of ints

```
def mysum(xs):
    return sum(x for x in xs)
```

It could be typed as `(List[int]) -> int` and be correct in the scope of the existing program but isn't it more extensible to type it as `(Iterable[int]) -> int`? (assuming that only iterables of ints are ever passed and not custom objects with `__radd__` impls). 

Similarly here, `Any` is a valid input based on the function's implementation, even though current callers only pass `Optional[str]`. We don't type super generically everywhere because it can make the code overly generic & hard to read but I think it makes sense to do it in places like this.

pyupgrade.py
mxr: I guess the test suite doesn't depend on this being a true bool then

pyupgrade.py

pyupgrade.py
mxr: it does not, the rename was enough

pyupgrade.py
mxr: Line length.  I can noqa it instead or make a sub-function

pyupgrade.py

pyupgrade.py
asottile: hmmm! can that line be changed to `conversion = ''` and still work?

pyupgrade.py
asottile: oh I meant more that `Parsed` and `Fmt` aren't quite clear what they're being used for

pyupgrade.py
mxr: Hmmmm the cast here was based on the current test suite. I can add a test and do the right thing :tm:

pyupgrade.py

pyupgrade.py
mxr: Gotcha

pyupgrade.py
mxr: I guess it passed tests? I haven't really tried to understand the code as well I could have 🙈

pyupgrade.py
mxr: I had to do the None check in the caller so that `int` wouldn't complain. So maybe this can just be `str` after all.

Another option is to create a `maybe_int` function which returns the int if it can be done and some bogus int if it's not possible, and use that for comparisons. But that has other problems

pyupgrade.py
mxr: yes

pyupgrade.py
mxr: urgh let me think on this..

pyupgrade.py
mxr: sure. assert too?

pyupgrade.py
mxr: ok

pyupgrade.py
mxr: yea

mxr:
@asottile think I hit everything in the last review 📿

pyupgrade.py
```diff
@@ -22,11 +36,31 @@
...
+    Optional[str],
+    str,
+]
+PercentFormat = Tuple[str, Optional[PercentFormatPart]]
+
+ListCompOrGeneratorExp = Union[ast.ListComp, ast.GeneratorExp]
+ListOrTuple = Union[ast.List, ast.Tuple]
+NameOrAttr = Union[ast.Name, ast.Attribute]
+
+if False:
```
asottile:
```suggestion
if False:  # pragma: no cover (mypy)
```

weird, maybe I'm wrong

pyupgrade.py
```diff
@@ -1186,15 +1254,23 @@ def _eq(test, n):  # type: (ast.Compare, int) -> bool
...
+        if not (
+                isinstance(test.ops[0], op) and
+                isinstance(test.comparators[0], ast.Tuple) and
+                len(test.comparators[0].elts) >= 1 and
+                all(isinstance(n, ast.Num) for n in test.comparators[0].elts)
+        ):
+            return False
+
+        # checked above but mypy needs help
+        elts = cast("List[ast.Num]", test.comparators[0].elts)
```
asottile: heh, I guess I don't have `double-quote-string-fixer` pre-commit hook enforced do I 🤔

setup.cfg
```diff
@@ -23,7 +23,9 @@ classifiers =
 
 [options]
 py_modules = pyupgrade
-install_requires = tokenize-rt>=3.0.1
+install_requires =
+    tokenize-rt>=3.0.1
+    typing; python_version=="2.7"
```
asottile: can you run `setup-cfg-fmt` on this, it should bump `python_requires` to `>= 3.4` so that we don't need to also install typing for that

mxr: I opened a new PR for it

pyupgrade.py
mxr: I opened another PR for that

pyupgrade.py
mxr: Want it in the coveragerc instead?

mxr: Hmm locally I still got 100% coverage even without this

mxr: Of course, locally doesn't matter ;)

asottile: python2 doesn't seem to get 100% coverage though and I'm not sure why -- might be the peephole optimizer

asottile: and I'd prefer here, I plan to blow away `.coveragerc` in all by repos with a "standard" copy instead of the current drift I'm maintaining

pyupgrade.py
mxr: Hmm I don't understand what you mean 🤣

asottile: ignore this, I'm babbling to myself -- I'll fix it some other day

pyupgrade.py

mxr: Maybe `coverage` is panic routing?

PR #180, opened 2019-07-05
stdin support

asottile:


asottile:
CC @theevocater since you didn't get to this one

PR #184, opened 2019-07-07
Fix type annotation

mxr:


mxr:
@asottile do you know why mypy didn't flag this? Is `Offset` correctly being read out from `tokenize_rt`?

asottile:
it's probably just defaulting to `Any` due to not having `tokenize-rt` installed if I had to guess  -- not even sure if it would find it even with it installed -- not even sure if it would work because PEP 561 doesn't work for single-file distributions

PR #185, opened 2019-07-07
Rewrite more OSError aliases

mxr:
More for #141.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,11 +291,23 @@ Availability:
 
 ```python
 # input
-x = EnvironmentError('boom')        # also understands IOError and WindowsError
-raise EnvironmentError('boom')
+
+# also understands:
+# - IOError
+# - WindowsError
+# - mmap.error and uses of `from mmap import error`
+# - select.error and uses of `from select import error`
+# - socket.error and uses of `from socket import error`
+
+try:
+    raise EnvironmentError('boom')
+except EnvironmentError:
+    raise
 # output
-x = OSError('boom')
-raise OSError('boom')
+try:
+    raise OSError('boom')
+except OSError:
+    raise
 ```
 
 ### f-strings
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1087,6 +1087,14 @@ class FindPy3Plus(ast.NodeVisitor):
         'WindowsError',
     ))
 
+    OS_ERROR_ALIAS_MODULES = frozenset((
+        'mmap',
+        'select',
+        'socket',
+    ))
+
+    FROM_IMPORTED_MODULES = OS_ERROR_ALIAS_MODULES.union(('six',))
+
     class ClassInfo:
         def __init__(self, name):  # type: (str) -> None
             self.name = name
@@ -1103,9 +1111,13 @@ class FindPy3Plus(ast.NodeVisitor):
         self.if_py3_blocks = set()  # type: Set[ast.If]
 
         self.native_literals = set()  # type: Set[Offset]
-        self.simple_ids = {}  # type: Dict[Offset, ast.Name]
 
-        self._six_from_imports = set()  # type: Set[str]
+        self._from_imported_names = collections.defaultdict(
+            set,
+        )  # type: Dict[str, Set[str]]
+        self.os_error_alias_calls = {}  # type: Dict[Offset, ast.Call]
+        self.os_error_alias_simple = {}  # type: Dict[Offset, NameOrAttr]
+
         self.six_add_metaclass = set()  # type: Set[ast.ClassDef]
         self.six_b = set()  # type: Set[ast.Call]
         self.six_calls = {}  # type: Dict[Offset, ast.Call]
@@ -1127,12 +1139,32 @@ class FindPy3Plus(ast.NodeVisitor):
         return (
             isinstance(node, ast.Name) and
             node.id in names and
-            node.id in self._six_from_imports
+            node.id in self._from_imported_names['six']
         ) or (
             isinstance(node, ast.Attribute) and
             isinstance(node.value, ast.Name) and
             node.value.id == 'six' and
             node.attr in names
+        )
+
+    def _is_os_error_alias(self, node):
+        # type: (ast.expr) -> bool
+        return (
+            isinstance(node, ast.Name) and
+            node.id in self.OS_ERROR_ALIASES
+        ) or (
+            isinstance(node, ast.Name) and
+            node.id == 'error' and
+            (
+                node.id in self._from_imported_names['mmap'] or
+                node.id in self._from_imported_names['select'] or
+                node.id in self._from_imported_names['socket']
+            )
+        ) or (
+            isinstance(node, ast.Attribute) and
+            isinstance(node.value, ast.Name) and
+            node.value.id in self.OS_ERROR_ALIAS_MODULES and
+            node.attr == 'error'
         )
 
     def _is_version_info(self, node):  # type: (ast.expr) -> bool
@@ -1148,10 +1180,10 @@ class FindPy3Plus(ast.NodeVisitor):
         )
 
     def visit_ImportFrom(self, node):  # type: (ast.ImportFrom) -> None
-        if node.module == 'six':
+        if node.module in self.FROM_IMPORTED_MODULES:
             for name in node.names:
                 if not name.asname:
-                    self._six_from_imports.add(name.name)
+                    self._from_imported_names[node.module].add(name.name)
         elif node.module == 'sys' and any(
             name.name == 'version_info' and not name.asname
             for name in node.names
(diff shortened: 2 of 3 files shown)


pyupgrade.py
```diff
@@ -1127,14 +1144,35 @@ def _is_six(self, node, names):
...
+        # type: (ast.Call) -> bool
+        if _starargs(node):
+            return False
+
+        func = node.func
+
+        return (
+            isinstance(func, ast.Name) and
+            func.id == 'error' and
+            any(
```
mxr: If I change this to 

```diff
-            any(
-                func.id in self._from_imported_names[modname]
-                for modname in self.OS_ERROR_ALIAS_MODULES
+            (
+                func.id in self._from_imported_names['mmap'] or
+                func.id in self._from_imported_names['select'] or
+                func.id in self._from_imported_names['socket']
             )
```

coverage passes.

Otherwise it fails saying that lines 1162 and beyond are uncovered (?)

mxr: Potentially related to `https://github.com/nedbat/coveragepy/issues/515` but I don't have proof..

mxr:
@asottile mind taking a look?

pyupgrade.py
```diff
@@ -1087,6 +1090,16 @@ class FindPy3Plus(ast.NodeVisitor):
...
     ))
 
+    OS_ERROR_ALIAS_MODULES = frozenset((
+        'mmap',
+        'select',
+        'socket',
+    ))
+
+    FROM_IMPORTED_MODULES = frozenset(
+        itertools.chain(('six',), OS_ERROR_ALIAS_MODULES),
```
asottile: no need for `chain`, just use `{'six'} | OS_ERROR_ALIAS_MODULES`

asottile: or `OS_ERROR_ALIAS_MODULES.union(('six',))`

pyupgrade.py
```diff
@@ -1103,9 +1116,13 @@ def __init__(self):  # type: () -> None
         self.if_py3_blocks = set()  # type: Set[ast.If]
 
         self.native_literals = set()  # type: Set[Offset]
-        self.simple_ids = {}  # type: Dict[Offset, ast.Name]
 
-        self._six_from_imports = set()  # type: Set[str]
+        self._from_imported_names = collections.defaultdict(
+            set,
+        )  # type: DefaultDict[str, Set[str]]
```
asottile: I've just been using `Dict` for this alias -- works well enough and then you don't need a special typing import for it (and we don't use any of the apis of `DefaultDict`)

pyupgrade.py
asottile: another idea is to ditch these constants entirely and record all the `from` imports

pyupgrade.py
```diff
@@ -1127,14 +1144,36 @@ def _is_six(self, node, names):
...
         ) or (
             isinstance(node, ast.Attribute) and
             isinstance(node.value, ast.Name) and
             node.value.id == 'six' and
             node.attr in names
         )
 
+    def _is_oserror_complex_candidate(self, node):
+        # type: (ast.Call) -> bool
+        if _starargs(node):
```
asottile: it's actually probably fine to allow starargs here -- we aren't rewriting any of the arguments

pyupgrade.py
```diff
@@ -1789,7 +1831,11 @@ def _replace(i, mapping, node):
             if any(tok.name == 'NL' for tok in tokens[i:end]):
                 continue
             _replace_call(tokens, i, end, func_args, '{args[0]}')
-        elif token.offset in visitor.simple_ids:
+        elif token.offset in visitor.oserror_complex:
+            j = _find_open_paren(tokens, i)
+            line, offset = token.line, token.utf8_byte_offset
+            tokens[i:j] = [Token('NAME', 'OSError', line, offset)]
```
asottile: can use `[token._replace(name='NAME', src='OSError')]`

tests/oserror_aliases_test.py
```diff
@@ -8,6 +8,26 @@
...
+    (
+        # if "error" isn't a function call it shouldn't be rewritten
+        # to avoid false positives
+        'from {imp} import error\n\n'
+        'def foo():\n'
+        '     error = 3\n',
+        '     return error\n',
+        # renaming things for weird reasons
+        'from {imp} import error as the_roof\n'
+        'raise the_roof()\n',
```
asottile: :rofl:

tests/oserror_aliases_test.py
```diff
@@ -8,6 +8,26 @@
 from pyupgrade import FindPy3Plus
 
 
+@pytest.mark.parametrize('imp', FindPy3Plus.OS_ERROR_ALIAS_MODULES)
+@pytest.mark.parametrize(
+    'tpl',
+    (
+        # if "error" isn't a function call it shouldn't be rewritten
```
asottile: I think we should handle `except` blocks -- they're far more likely than construction of these error classes imo

mxr: More generally, should I change the OSError fixer to only change `raise` and `except` blocks?

asottile: yeah probably

pyupgrade.py
mxr: will do the union approach for now just to be slightly more DRY

pyupgrade.py

pyupgrade.py

pyupgrade.py

tests/oserror_aliases_test.py
mxr: done

mxr:
@asottile this is ready for another round

asottile:
> @asottile this is ready for another round

lol python2 :'(

mxr:
EOL when

mxr:
image

asottile:
> Raise has no attribute 'exc'

\*cries in python 2\*

mxr:
Should be good under py2 (tested locally though we've seen how much good that is)

pyupgrade.py
```diff
@@ -1227,12 +1259,43 @@ def _visit_simple(self, node):  # type: (NameOrAttr) -> None
...
+    def visit_Try(self, node):  # type: (ast.Try) -> None
+        for handler in node.handlers:
+            htype = handler.type
+            if not htype:
+                continue
+            elif self._is_os_error_alias(htype):
+                assert isinstance(htype, (ast.Name, ast.Attribute))
+                self.os_error_alias_simple[_ast_to_offset(htype)] = htype
+            else:
+                assert isinstance(htype, ast.Tuple)
```
asottile:
```python
try:
    pass
except AssertionError:
    pass
```

```console
$ pyupgrade t.py  --py3-plus
  File "/home/asottile/workspace/pyupgrade/venv/bin/pyupgrade", line 11, in <module>
    load_entry_point('pyupgrade', 'console_scripts', 'pyupgrade')()
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 2062, in main
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 2031, in fix_file
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1708, in _fix_py3_plus
    visitor.visit(ast_obj)
  File "/usr/lib/python3.6/ast.py", line 253, in visit
    return visitor(node)
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1443, in generic_visit
  File "/usr/lib/python3.6/ast.py", line 261, in generic_visit
    self.visit(item)
  File "/usr/lib/python3.6/ast.py", line 253, in visit
    return visitor(node)
  File "/home/asottile/workspace/pyupgrade/pyupgrade.py", line 1273, in visit_Try
    assert isinstance(htype, ast.Tuple)
AssertionError
```

pyupgrade.py
```diff
@@ -1789,13 +1857,15 @@ def _replace(i, mapping, node):
...
-        elif token.offset in visitor.simple_ids:
-            tokens[i] = Token(
-                token.name,
-                'OSError',
-                token.line,
-                token.utf8_byte_offset,
-            )
+        elif token.offset in visitor.os_error_alias_calls:
+            node = visitor.os_error_alias_calls[token.offset]
+            if isinstance(node.func, ast.Name):
```
asottile: hmmm you end up with `Name`s in the `_calls` mapping? shouldn't that be in the `alias_simple` mapping instead

pyupgrade.py
asottile: also

```python
try:
    pass
except [AssertionError] as e:
    pass
```

```python
try:
    pass
except {AssertionError} as e:
    pass
```

```python
def hello_world():
    return AssertionError

try:
    pass
except hello_world() as e:
    pass
```

pyupgrade.py
mxr: I think I end up parsing the `Name` that's inside the `Call` :/ will think about this some more

pyupgrade.py
mxr: makes sense 🤦‍♂ this assertion was to work around coverage

```diff
-            else:
-                assert isinstance(htype, ast.Tuple)
+            elif isinstance(htype, ast.Tuple):
```

image

I'll work around this another way

mxr: Adding that test case and restructuring the code a bit did the trick 👍

pyupgrade.py
mxr: err that's not it

```diff
             if isinstance(node.func, ast.Name):
+                import ipdb; ipdb.set_trace()
                 _replace_os_error(i, node.func)
```

```
ipdb> import astpretty                                                                                                                                            
ipdb> astpretty.pprint(node)                                                                                                                                      
Call(
    lineno=1,
    col_offset=6,
    func=Name(lineno=1, col_offset=6, id='EnvironmentError', ctx=Load()),
    args=[
        Num(lineno=1, col_offset=23, n=1),
        Num(lineno=1, col_offset=26, n=2),
    ],
    keywords=[],
)
```

This is triggered by the `main` test which is just `raise EnvironmentError(1, 2)`.

pyupgrade.py
asottile: just means you need more tests :P

pyupgrade.py
```diff
@@ -1728,16 +1789,16 @@ def _replace(i, mapping, node):
...
+            assert isinstance(call.func, (ast.Name, ast.Attribute))
+            template = _get_tmpl(SIX_CALLS, call.func)
             _replace_call(tokens, i, end, func_args, template)
         elif token.offset in visitor.six_raises:
             j = _find_open_paren(tokens, i)
             func_args, end = _parse_call_args(tokens, j)
-            node = visitor.six_raises[token.offset]
-            assert isinstance(node.func, (ast.Name, ast.Attribute))
-            template = _get_tmpl(SIX_RAISES, node.func)
+            call = visitor.six_raises[token.offset]
```
mxr: This was for mypy (so I could use the `node` variable below

README.md
```diff
@@ -291,11 +291,24 @@ Availability:
 
 ```python
 # input
-x = EnvironmentError('boom')        # also understands IOError and WindowsError
-raise EnvironmentError('boom')
+def f():
```
asottile: I don't think you need a wrapping function here, right?

mxr: nope. I can remove it

mxr:
@asottile 	updated the readme

PR #191, opened 2019-07-12
Rewrite `str()` as `''`

### LINKED ISSUE #187: `str()` native literal should be just `''`
https://github.com/asottile/pyupgrade#forced-strnative-literals

```python
str()  # ''
```

mxr:
Resolves #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ Availability:
 - `--py3-plus` is passed on the commandline.
 
 ```python
+str()       # "''"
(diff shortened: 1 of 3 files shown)


mxr:
A `--black=true` option to do the rewrite as `'""'` instead is an exercise left to the reader

asottile:
> A `--black=true` option to do the rewrite as `'""'` instead is an exercise left to the reader

lol nty

PR #193, opened 2019-07-22
Rewrite multiple OSError aliases in the same `except` block

mxr:
Brought up in #141

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1112,12 +1112,13 @@ class FindPy3Plus(ast.NodeVisitor):
 
         self.native_literals = set()  # type: Set[Offset]
 
-        self._from_imported_names = collections.defaultdict(
+        self.from_imported_names = collections.defaultdict(
             set,
         )  # type: Dict[str, Set[str]]
         self.io_open_calls = {}  # type: Dict[Offset, ast.Call]
         self.os_error_alias_calls = {}  # type: Dict[Offset, ast.Call]
         self.os_error_alias_simple = {}  # type: Dict[Offset, NameOrAttr]
+        self.os_error_alias_excepts = {}  # type: Dict[Offset, ast.Tuple]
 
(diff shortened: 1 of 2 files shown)


pyupgrade.py
```diff
@@ -1880,6 +1889,70 @@ def _replace(i, mapping, node):
...
+                        right == 'error'
+                ):
+                    if not seen_os_error:
+                        args.append('OSError')
+                    seen_os_error = True
+                elif (
+                        left in visitor.OS_ERROR_ALIASES and
+                        part == right == ''
+                ):
+                    if not seen_os_error:
```
asottile: could this be simplified with a `set` and `list`?

tests/oserror_aliases_test.py
```diff
@@ -8,6 +8,72 @@
...
+            '    pass\n',
+
+            'try:\n'
+            '    pass\n'
+            'except OSError:\n'
+            '    pass\n',
+        ),
+    ),
+)
+def test_fix_oserror_aliases_try(alias, tpl, expected):
```
asottile: can you add a test where the tuple spans multiple lines?

pyupgrade.py
mxr: I wanted to keep the order, will use `list(OrderedDict.from_keys(...))`

tests/oserror_aliases_test.py
mxr: sure

mxr:
pushed up 🙏

asottile:
can you add this test in the noop category:

```python
try:
    pass
except (
    T,
):
    pass
```

Also maybe split the formatted tests out from the non-formatted tests -- every non-formatted test added to the noop test adds 3 identical tests

mxr:
Sure. I added the tests, added some test cases with multiple excepts, and reorganized the code a bit.

PR #196, opened 2019-08-08
moved string formatting into its own function

verhovsky:
I copied [blacks function name and argument](https://github.com/psf/black/blob/bd0ab3cba2fb2be65985741f0a4c382b675c2876/black.py#L707), since the fact that it's `_text` (as opposed to `contents_bytes`) is a given for this function

See https://github.com/verhovsky/pyupgrade-docs/pull/1


asottile:
I don't want to support this, sorry

PR #208, opened 2019-10-07
Special-case `next(six.iter*())` rewrite

### LINKED ISSUE #206: next(six.itervalues(d)) is rewritten to next(d.values())
```console
$ cat example.py
import six


print(next(six.itervalues({1: 2})))
$ python example.py
2
```
```console
$ pyupgrade --py36-plus example.py
Rewriting example.py
```
```console
$ cat example.py
import six


print(next({1: 2}.values()))
$ python example.py
  File "example.py", line 4, in <module>
    print(next({1: 2}.values()))
TypeError: 'dict_values' object is not an iterator
```

mxr:
Resolves #206

### DIFF
--- a/tests/six_test.py
+++ b/tests/six_test.py
@@ -338,15 +338,14 @@ def test_fix_six_noop(s):
             id='add_metaclass, indented',
         ),
         pytest.param(
-            'print(six.itervalues({1:2}))\n'
-            'print(next(six.itervalues({1:2})))\n'
             'print(six.itervalues({1:2}))\n',
-
-            'print({1:2}.values())\n'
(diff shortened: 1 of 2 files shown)


tests/six_test.py
```diff
@@ -337,6 +337,17 @@ def test_fix_six_noop(s):
 
             id='add_metaclass, indented',
         ),
+        pytest.param(
+            'print(six.itervalues({1:2}))\n'
+            'print(next(six.itervalues({1:2})))\n'
+            'print(six.itervalues({1:2}))\n',
```
asottile: you've got a dupe test here

mxr:
+1

PR #218, opened 2019-10-21
Fix type annotations

atugushev:


atugushev:
@asottile, thanks for merging this!

PR #220, opened 2019-11-01
Fix unsafe yield from rewrite

### LINKED ISSUE #216: 'yield from' rewrite unsafe when looping variable is referenced later
Consider the example code:

```py
def foo(it):
    item = None
    for item in it:
        yield item
    print("Last item", item)
```

With `--py3-plus`, this is transformed to:

```py
def foo(it):
    item = None
    yield from it
    print("Last item", item)
```

But these are not equivalent. The line `print("Last item", item)` now always prints `None` rather than the last item of the passed iterable. 

This was discovered in more-itertools:

https://github.com/erikrose/more-itertools/blob/8d860c355a1dd900e3a9c5fe1303ddf5a778452f/more_itertools/more.py#L1206-L1223

atugushev:
Fixes #216.

Prevents rewrite `for-loop yield` if a loop variable referenced after the loop.

Also, it fixes along the way the case with yield outside the function. Now `pyupgrade` doesn't rewrite it:
```bash
# before
$ echo "for x in y: yield x" | pyupgrade - --py3-plus
yield from y

# after
$ echo "for x in y: yield x" | pyupgrade - --py3-plus
for x in y: yield x
```
<sup>I don't know is it intentional behavior or not. Please, inform if it is.</sup>

Additionally, I've added a test for `lambda yield` (weird but syntactically possible), since there is `Union[ast.FunctionDef, ast.Lambda]`.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1557,7 +1557,7 @@ class FindNamesRefsAfterForLoopVisitor(ast.NodeVisitor):
         self.generic_visit(node)
 
     def visit(self, node):  # type: (ast.AST) -> None
-        super().visit(node)
+        super(FindNamesRefsAfterForLoopVisitor, self).visit(node)
         self._stack_depth += 1
 
 
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -51,6 +51,7 @@ PercentFormat = Tuple[str, Optional[PercentFormatPart]]
 ListCompOrGeneratorExp = Union[ast.ListComp, ast.GeneratorExp]
 ListOrTuple = Union[ast.List, ast.Tuple]
 NameOrAttr = Union[ast.Name, ast.Attribute]
+SyncFunctionDef = Union[ast.FunctionDef, ast.Lambda]
 
 if False:  # pragma: no cover (mypy)
     from typing import Type
@@ -1196,7 +1197,7 @@ class FindPy3Plus(ast.NodeVisitor):
         self._class_info_stack = []  # type: List[FindPy3Plus.ClassInfo]
         self._in_comp = 0
         self.super_calls = {}  # type: Dict[Offset, ast.Call]
-        self._in_async_def = False
+        self._current_sync_def = None  # type: Optional[SyncFunctionDef]
         self.yield_from_fors = set()  # type: Set[Offset]
 
     def _is_six(self, node, names):
@@ -1307,18 +1308,12 @@ class FindPy3Plus(ast.NodeVisitor):
             self.generic_visit(node)
 
     def _visit_sync_func(self, node):
-        # type: (Union[ast.FunctionDef, ast.Lambda]) -> None
-        self._in_async_def, orig = False, self._in_async_def
+        # type: (SyncFunctionDef) -> None
+        self._current_sync_def, orig = node, self._current_sync_def
         self._visit_func(node)
-        self._in_async_def = orig
+        self._current_sync_def = orig
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
-
-    def visit_AsyncFunctionDef(self, node):  # pragma: no cover (py35+)
-        # type: (AsyncFunctionDef) -> None
-        self._in_async_def, orig = True, self._in_async_def
-        self._visit_func(node)
-        self._in_async_def = orig
 
     def _visit_comp(self, node):  # type: (ast.expr) -> None
         self._in_comp += 1
@@ -1520,12 +1515,15 @@ class FindPy3Plus(ast.NodeVisitor):
 
     def visit_For(self, node):  # type: (ast.For) -> None
         if (
-            not self._in_async_def and
+            self._current_sync_def is not None and
             len(node.body) == 1 and
             isinstance(node.body[0], ast.Expr) and
             isinstance(node.body[0].value, ast.Yield) and
             node.body[0].value.value is not None and
             targets_same(node.target, node.body[0].value.value) and
+            not targets_referenced_after_node(
+                self._current_sync_def, node, node.target,
+            ) and
             not node.orelse
         ):
             self.yield_from_fors.add(_ast_to_offset(node))
@@ -1535,6 +1533,37 @@ class FindPy3Plus(ast.NodeVisitor):
     def generic_visit(self, node):  # type: (ast.AST) -> None
         self._previous_node = node
         super(FindPy3Plus, self).generic_visit(node)
+
+
+def reversed_walk(node):  # type: (ast.AST) -> Generator[ast.AST, None, None]
+    todo = [node]
+    while todo:
+        node = todo.pop()
+        todo.extend(ast.iter_child_nodes(node))
+        yield node
+
+
+def targets_referenced_after_node(root_node, node, targets):
+    # type: (ast.AST, ast.AST, Union[ast.expr, ast.Tuple]) -> bool
+    """
+    Search inside the given root node whether the targets have been referenced
+    after the given node. Walks from behind until the node have been reached.
+    """
+    for x in reversed_walk(root_node):
+        # Finish walking when the node has been reached
+        if x == node:
+            break
+
+        # Filter for name expressions
+        if not isinstance(x, ast.Name):
+            continue
+
+        # Match one of the targets
+        for t in ast.walk(targets):
+            if fields_same(t, x):
+                return True
+
+    return False
 
 
 def _fixup_dedent_tokens(tokens):  # type: (List[Token]) -> None
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1197,7 +1197,7 @@ class FindPy3Plus(ast.NodeVisitor):
         self._class_info_stack = []  # type: List[FindPy3Plus.ClassInfo]
         self._in_comp = 0
         self.super_calls = {}  # type: Dict[Offset, ast.Call]
-        self._current_sync_def = None  # type: Optional[SyncFunctionDef]
+        self._current_func_node = None  # type: Optional[SyncFunctionDef]
         self.yield_from_fors = set()  # type: Set[Offset]
 
     def _is_six(self, node, names):
@@ -1309,9 +1309,9 @@ class FindPy3Plus(ast.NodeVisitor):
 
     def _visit_sync_func(self, node):
         # type: (SyncFunctionDef) -> None
-        self._current_sync_def, orig = node, self._current_sync_def
+        self._current_func_node, orig = node, self._current_func_node
         self._visit_func(node)
-        self._current_sync_def = orig
+        self._current_func_node = orig
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
 
@@ -1515,15 +1515,13 @@ class FindPy3Plus(ast.NodeVisitor):
 
     def visit_For(self, node):  # type: (ast.For) -> None
         if (
-            self._current_sync_def is not None and
+            self._current_func_node is not None and
             len(node.body) == 1 and
             isinstance(node.body[0], ast.Expr) and
             isinstance(node.body[0].value, ast.Yield) and
             node.body[0].value.value is not None and
             targets_same(node.target, node.body[0].value.value) and
-            not targets_referenced_after_node(
-                self._current_sync_def, node, node.target,
-            ) and
+            not _targets_referenced(self._current_func_node, node) and
             not node.orelse
         ):
             self.yield_from_fors.add(_ast_to_offset(node))
@@ -1535,35 +1533,46 @@ class FindPy3Plus(ast.NodeVisitor):
         super(FindPy3Plus, self).generic_visit(node)
 
 
-def reversed_walk(node):  # type: (ast.AST) -> Generator[ast.AST, None, None]
-    todo = [node]
-    while todo:
-        node = todo.pop()
-        todo.extend(ast.iter_child_nodes(node))
-        yield node
+class FindNamesRefsAfterForLoopVisitor(ast.NodeVisitor):
+    def __init__(self, forloop):  # type: (ast.For) -> None
+        self.forloop = forloop
+
+        self._stack_depth = 0
+        self._forloop_depth = None  # type: Optional[int]
+        self.found = set()  # type: Set[str]
+
+    def visit_For(self, node):  # type: (ast.For) -> None
+        if node == self.forloop:
+            self._forloop_depth = self._stack_depth
+        else:
+            self.generic_visit(node)
+
+    def visit_Name(self, node):  # type: (ast.Name) -> None
+        if (
+            isinstance(node.ctx, ast.Load) and
+            self._forloop_depth is not None and
+            self._stack_depth > self._forloop_depth
+        ):
+            self.found.add(node.id)
+        self.generic_visit(node)
+
+    def visit(self, node):  # type: (ast.AST) -> None
+        super().visit(node)
+        self._stack_depth += 1
 
 
-def targets_referenced_after_node(root_node, node, targets):
-    # type: (ast.AST, ast.AST, Union[ast.expr, ast.Tuple]) -> bool
+def _targets_referenced(func, forloop):
+    # type: (SyncFunctionDef, ast.For) -> bool
+    """Search inside the given function node whether the loop variables
+    have been referenced after the loop.
     """
-    Search inside the given root node whether the targets have been referenced
-    after the given node. Walks from behind until the node have been reached.
-    """
-    for x in reversed_walk(root_node):
-        # Finish walking when the node has been reached
-        if x == node:
-            break
-
-        # Filter for name expressions
-        if not isinstance(x, ast.Name):
-            continue
-
-        # Match one of the targets
-        for t in ast.walk(targets):
-            if fields_same(t, x):
-                return True
-
-    return False
+    visitor = FindNamesRefsAfterForLoopVisitor(forloop)
+    visitor.visit(func)
+    names = {
+        node.id for node in ast.walk(forloop.target)
+        if isinstance(node, ast.Name)
+    }
+    return len(visitor.found & names) > 0
 
 
 def _fixup_dedent_tokens(tokens):  # type: (List[Token]) -> None
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1315,6 +1315,10 @@ class FindPy3Plus(ast.NodeVisitor):
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
 
+    def visit_AsyncFunctionDef(self, node):  # pragma: no cover (py35+)
+        # type: (AsyncFunctionDef) -> None
+        self._visit_func(node)
+
     def _visit_comp(self, node):  # type: (ast.expr) -> None
         self._in_comp += 1
         self.generic_visit(node)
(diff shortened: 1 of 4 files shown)


atugushev:
Aw, sorry. I've forgotten to check the coverage.

pyupgrade.py
```diff
@@ -1537,6 +1535,37 @@ def generic_visit(self, node):  # type: (ast.AST) -> None
         super(FindPy3Plus, self).generic_visit(node)
 
 
+def reversed_walk(node):  # type: (ast.AST) -> Generator[ast.AST, None, None]
+    todo = [node]
+    while todo:
```
atugushev: @asottile would it be okay to add `pragma: no branch` (for it's `while True`) here and further walk-branches? Or it should be refactored?

pyupgrade.py
```diff
@@ -1307,19 +1308,13 @@ def _visit_func(self, node):
...
         self._visit_func(node)
-        self._in_async_def = orig
+        self._current_sync_def = orig
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
 
-    def visit_AsyncFunctionDef(self, node):  # pragma: no cover (py35+)
-        # type: (AsyncFunctionDef) -> None
-        self._in_async_def, orig = True, self._in_async_def
-        self._visit_func(node)
```
asottile: this introduces a bug, can you write a test for old `super` inside an `async def`?

pyupgrade.py
asottile: you should be able to hit the coverage here -- the missing branch indicates this either never terminates or never loops more than once

pyupgrade.py
```diff
@@ -1537,6 +1535,37 @@ def generic_visit(self, node):  # type: (ast.AST) -> None
...
+    while todo:
+        node = todo.pop()
+        todo.extend(ast.iter_child_nodes(node))
+        yield node
+
+
+def targets_referenced_after_node(root_node, node, targets):
+    # type: (ast.AST, ast.AST, Union[ast.expr, ast.Tuple]) -> bool
+    """
+    Search inside the given root node whether the targets have been referenced
```
asottile: docstrings should not start with whitespace

pyupgrade.py
```diff
@@ -1537,6 +1535,37 @@ def generic_visit(self, node):  # type: (ast.AST) -> None
...
 
+def reversed_walk(node):  # type: (ast.AST) -> Generator[ast.AST, None, None]
+    todo = [node]
+    while todo:
+        node = todo.pop()
+        todo.extend(ast.iter_child_nodes(node))
+        yield node
+
+
+def targets_referenced_after_node(root_node, node, targets):
```
asottile: let's underscore these functions, pyupgrade has no programmatic api

pyupgrade.py
```diff
@@ -1537,6 +1535,37 @@ def generic_visit(self, node):  # type: (ast.AST) -> None
...
+    # type: (ast.AST, ast.AST, Union[ast.expr, ast.Tuple]) -> bool
+    """
+    Search inside the given root node whether the targets have been referenced
+    after the given node. Walks from behind until the node have been reached.
+    """
+    for x in reversed_walk(root_node):
```
asottile: alternatively you forward walk (don't need the other function) and then you start looking when the node is encountered

pyupgrade.py
```diff
@@ -1520,12 +1515,15 @@ def visit_If(self, node):  # type: (ast.If) -> None
...
-            not self._in_async_def and
+            self._current_sync_def is not None and
             len(node.body) == 1 and
             isinstance(node.body[0], ast.Expr) and
             isinstance(node.body[0].value, ast.Yield) and
             node.body[0].value.value is not None and
             targets_same(node.target, node.body[0].value.value) and
+            not targets_referenced_after_node(
+                self._current_sync_def, node, node.target,
+            ) and
```
asottile: not a big deal but if you could shorten this to fit on one line that'd be great

tests/yield_from_test.py
```diff
@@ -184,6 +184,48 @@ def test_fix_async_yield_from(s, expected):
         '        yield x\n'
         '    else:\n'
         '        print("boom!")\n',
+        pytest.param(
+            'for x in z:\n'
+            '    yield x\n',
+            id='yield outside function',
+        ),
+        pytest.param(
+            'lambda: [(yield x) for x in z]',
```
asottile: I don't think this test is relevant, there's no `for` loop here

pyupgrade.py
```diff
@@ -1537,6 +1535,37 @@ def generic_visit(self, node):  # type: (ast.AST) -> None
...
+            break
+
+        # Filter for name expressions
+        if not isinstance(x, ast.Name):
+            continue
+
+        # Match one of the targets
+        for t in ast.walk(targets):
+            if fields_same(t, x):
+                return True
```
asottile: It would probably be better to refactor to this (above the loop)

```python
names = {node.id for node in ast.walk(targets)}

...

if isinstance(x, ast.Name) and isinstance(x.ctx, ast.Load) and x.id in names:
    ...
```

atugushev:
```python
names = {node.id for node in ast.walk(targets)}
```
That's neat, but it seems it doesn't handle targets as `ast.Tuple` (`for x, y in z: yield x, y` case). That's why there's `fields_same`.

atugushev: Maybe, it's better to walk through the targets and compose the set with node.id.

asottile: it's missing an `if isinstance(node, ast.Name)` -- that's what I get for phone

atugushev: Yay. that's working!

pyupgrade.py
atugushev: The naming and signature are okay?

asottile: yeah I meant if you can come up with more succinct names such that it fits, not a big deal

pyupgrade.py
atugushev: I came up with:
```python
todo = collections.deqeu(func_node.body)
while todo:
    node = todo.popleft()
    ....
```
What do you think?

asottile: I'd prefer trying to make it work with `ast.walk` if possible -- failing that whatever works

atugushev: Can't figure out how to achieve the same with `ast.walk`

```python
    visited = False
    todo = collections.deque(func_node.body)

    while todo:
        node = todo.popleft()

        if node == forloop_node:
            visited = True
            continue

        if not visited:
            continue

        if (
            isinstance(node, ast.Name) and
            isinstance(node.ctx, ast.Load) and
            node.id in names
        ):
            return True

        todo.extend(ast.iter_child_nodes(node))

    return False
```

How to prevent walking inside the for loop? Have any ideas?

atugushev: Dammit, I'm too bad at walking trees 😒

asottile: adjust the `visited` assignment until the last element of the `yield` expression maybe

atugushev: Too complicated, maybe node visitor would help? I'll try that.

asottile: if you're using a node visitor you could reuse the existing walk and remove the loops that are referenced after -- this would also keep it 1-pass :) (you might need to store some additional state, but you could use `visit_Name` directly)

atugushev: Strange just happened. I wrote the visitor and tests passed :D And it's less code.

atugushev: Just a thought, can we skip the nodes till for loop depend on offset information?

asottile: not always, for instance

```python
f'''
{foo}
'''
```

reports an offset of -1 on python<3.8

pyupgrade.py
atugushev: Addressed in 81dc036f6a52e32a3dd1e992334c24fc09cea12b.

pyupgrade.py
atugushev: Addressed in 81dc036f6a52e32a3dd1e992334c24fc09cea12b.

pyupgrade.py
atugushev: Addressed in 81dc036f6a52e32a3dd1e992334c24fc09cea12b.

pyupgrade.py
atugushev: Please review in 81dc036f6a52e32a3dd1e992334c24fc09cea12b  how it's done via visitor and stack depth level checking.

pyupgrade.py
atugushev: Addressed in 81dc036f6a52e32a3dd1e992334c24fc09cea12b.

pyupgrade.py
atugushev: Addressed in 81dc036f6a52e32a3dd1e992334c24fc09cea12b.

tests/yield_from_test.py
atugushev: Addressed in 9572c5a7dadf6a4f8c5877462aa23d05be9168fc.

pyupgrade.py
atugushev: Sorry, I'm afraid I'm not following. Could you elaborate on this?

atugushev: Should I revert back this block and restore `if not self._in_async_def` in `visit_For`?

asottile: this fails rewrite now:

```python
class C:
    async def foo(self):
        super(C, self).foo()
```

just needs a quick test for that and to make sure this still calls `_visit_func` essentially

atugushev: I see. Thanks for noticing!

atugushev: Addressed in 302ab16283facc67a6cc5ab189af10ba915ae05b.

pyupgrade.py
asottile: cool I'll poke at this, might be another way to approach it I'll see what I can do 👍

atugushev: Thank you so much!

atugushev: > adjust the `visited` assignment until the last element of the `yield` expression maybe

FTR, looks like it's impossible since `ast.walk` does undefined-order traversal.

asottile:
can you add this testcase (should be a noop):

```python
def f():
    async def g():
        for y in z:
            yield y
    return g
```

asottile:
ok had another chance to look at this:

```diff
diff --git a/pyupgrade.py b/pyupgrade.py
index 8cd3a13..2f4ec67 100644
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1197,7 +1197,8 @@ class FindPy3Plus(ast.NodeVisitor):
         self._class_info_stack = []  # type: List[FindPy3Plus.ClassInfo]
         self._in_comp = 0
         self.super_calls = {}  # type: Dict[Offset, ast.Call]
-        self._current_func_node = None  # type: Optional[SyncFunctionDef]
+        self._current_func = None  # type: Optional[SyncFunctionDef]
+        self._for_targets = {}  # type: Dict[Tuple[SyncFunctionDef, str], Offset]  # noqa: E501
         self.yield_from_fors = set()  # type: Set[Offset]
 
     def _is_six(self, node, names):
@@ -1309,9 +1310,9 @@ class FindPy3Plus(ast.NodeVisitor):
 
     def _visit_sync_func(self, node):
         # type: (SyncFunctionDef) -> None
-        self._current_func_node, orig = node, self._current_func_node
+        self._current_func, orig = node, self._current_func
         self._visit_func(node)
-        self._current_func_node = orig
+        self._current_func = orig
 
     visit_FunctionDef = visit_Lambda = _visit_sync_func
 
@@ -1518,17 +1519,28 @@ class FindPy3Plus(ast.NodeVisitor):
         self.generic_visit(node)
 
     def visit_For(self, node):  # type: (ast.For) -> None
+        self.generic_visit(node)
+        # done *after* `generic_visit` so we know all things after this are
+        # after this `for` loop
         if (
-            self._current_func_node is not None and
+            self._current_func is not None and
             len(node.body) == 1 and
             isinstance(node.body[0], ast.Expr) and
             isinstance(node.body[0].value, ast.Yield) and
             node.body[0].value.value is not None and
             targets_same(node.target, node.body[0].value.value) and
-            not _targets_referenced(self._current_func_node, node) and
             not node.orelse
         ):
-            self.yield_from_fors.add(_ast_to_offset(node))
+            offset = _ast_to_offset(node)
+            self.yield_from_fors.add(offset)
+            for node in ast.walk(node.target):
+                if isinstance(node, ast.Name):
+                    self._for_targets[(self._current_func, node.id)] = offset
+
+    def visit_Name(self, node):  # type: (ast.Name) -> None
+        offset = self._for_targets.get((self._current_func, node.id))
+        if offset is not None:
+            self.yield_from_fors.discard(offset)
 
         self.generic_visit(node)
 
@@ -1537,48 +1549,6 @@ class FindPy3Plus(ast.NodeVisitor):
         super(FindPy3Plus, self).generic_visit(node)
 
 
-class FindNamesRefsAfterForLoopVisitor(ast.NodeVisitor):
-    def __init__(self, forloop):  # type: (ast.For) -> None
-        self.forloop = forloop
-
-        self._stack_depth = 0
-        self._forloop_depth = None  # type: Optional[int]
-        self.found = set()  # type: Set[str]
-
-    def visit_For(self, node):  # type: (ast.For) -> None
-        if node == self.forloop:
-            self._forloop_depth = self._stack_depth
-        else:
-            self.generic_visit(node)
-
-    def visit_Name(self, node):  # type: (ast.Name) -> None
-        if (
-            isinstance(node.ctx, ast.Load) and
-            self._forloop_depth is not None and
-            self._stack_depth > self._forloop_depth
-        ):
-            self.found.add(node.id)
-        self.generic_visit(node)
-
-    def visit(self, node):  # type: (ast.AST) -> None
-        super(FindNamesRefsAfterForLoopVisitor, self).visit(node)
-        self._stack_depth += 1
-
-
-def _targets_referenced(func, forloop):
-    # type: (SyncFunctionDef, ast.For) -> bool
-    """Search inside the given function node whether the loop variables
-    have been referenced after the loop.
-    """
-    visitor = FindNamesRefsAfterForLoopVisitor(forloop)
-    visitor.visit(func)
-    names = {
-        node.id for node in ast.walk(forloop.target)
-        if isinstance(node, ast.Name)
-    }
-    return len(visitor.found & names) > 0
-
-
 def _fixup_dedent_tokens(tokens):  # type: (List[Token]) -> None
     """For whatever reason the DEDENT / UNIMPORTANT_WS tokens are misordered
 
```

let me know what you think of that

also probably needs a test for:

```python
def f():
    for x in y:
        yield x
    x += 1
```

I think this uses `ctx=Store()` despite being an augmented store :S

atugushev:
> let me know what you think of that

That's brilliant!

atugushev:
I've applied your patch in 424410f and added tests and a few changes to make mypy happy in a083dcf.

pyupgrade.py
```diff
@@ -1295,7 +1301,8 @@ def visit_ClassDef(self, node):  # type: (ast.ClassDef) -> None
         self._class_info_stack.pop()
 
     def _visit_func(self, node):
-        # type: (Union[ast.FunctionDef, ast.Lambda, AsyncFunctionDef]) -> None
+        # type: (SyncOrAsyncFunctionDef) -> None
+        self._current_func, orig = node, self._current_func
```
asottile: instead you can probably set `_current_func` to `None` inside `visit_AsyncFunctionDef`

tests/super_test.py
```diff
@@ -110,6 +112,19 @@ def test_fix_super_noop(s):
...
+            '        super(C, self).foo()\n',
+            'class C:\n'
+            '    async def foo(self):\n'
+            '        super().foo()\n',
+            id='super inside async func def',
+            marks=pytest.mark.xfail(
+                sys.version_info < (3, 5),
+                reason='async introduced in python 3.5',
+            ),
+        ),
```
asottile: let's split this into a separate function

pyupgrade.py
```diff
@@ -58,6 +58,11 @@
         AsyncFunctionDef = ast.AsyncFunctionDef
     else:
         AsyncFunctionDef = ast.stmt
+    SyncOrAsyncFunctionDef = Union[
+        ast.AsyncFunctionDef,
+        ast.FunctionDef,
+        ast.Lambda
```
asottile: if you upgrade `add-trailing-comma` to the latest version it will fix this

pyupgrade.py
atugushev: Addressed in a070506d51ca8e715a83912e6adaa8f6fbcc0b85.

tests/super_test.py
atugushev: Addressed in 626fead0f4bba195eed39bc5ee9b1b9612db4bfc.

pyupgrade.py
atugushev: Addressed in 7f7709432965ef668110d1fe96f932b9f02d3a88.

tests/super_test.py
```diff
@@ -129,3 +129,25 @@ def test_fix_super_noop(s):
...
+            'class C:\n'
+            '    async def foo(self):\n'
+            '        super(C, self).foo()\n',
+            'class C:\n'
+            '    async def foo(self):\n'
+            '        super().foo()\n',
+            id='super inside async func def',
+        ),
+    ),
+)
```
asottile: probably don't need parametrize if there's only one test

pyupgrade.py
```diff
@@ -1196,7 +1201,8 @@ def __init__(self):  # type: () -> None
         self._class_info_stack = []  # type: List[FindPy3Plus.ClassInfo]
         self._in_comp = 0
         self.super_calls = {}  # type: Dict[Offset, ast.Call]
-        self._in_async_def = False
+        self._current_func = None  # type: Optional[SyncOrAsyncFunctionDef]
```
asottile: this type is now overly broad here (I'd probably remove `SyncOrAsyncFunctionDef` as well)

atugushev: Would you like to change it to
```python
self._current_func = None  # type: Optional[Union[ast.AsyncFunctionDef, ast.FunctionDef, ast.Lambda]]  # noqa: E501
```

instead?

tests/super_test.py
atugushev: Yeah, right. I've forgotten to remove the old test btw 🤦‍♂

pyupgrade.py
asottile: you changed it so it can no longer contain `AsyncFunctionDef`

asottile:
hmm shoot I just thought of another case :/

```python
def f():
    def g():
        return x

    for x in y:
        yield x
    yield g()
```

this complicates things significantly I think :/

pyupgrade.py
atugushev: Got it!

atugushev:
Ouch!

cat

asottile:
so the adjusted logic probably needs to be something like: accessed anywhere outside of the loop body :/

probably needs a stack of nested functions and to store referenced names in them (and avoid it in the loop body)

this is much more complicated than initially thought lol

tests/super_test.py
atugushev: Addressed in 56e91882b1c626d72c97882f0352d21ae93ae0fb.

pyupgrade.py
atugushev: Addressed in 56e91882b1c626d72c97882f0352d21ae93ae0fb.

asottile:
via #286

Issue #219, opened 2019-11-01
Delete empty lines and empty comment lines after "coding" comment


verhovsky:
Stuff like this

```python
# -*- coding: utf-8 -*-
#  
# This is a Python script that does foo.
```

gets turned into

```python
# 
# This is a Python script that does foo.
```

`pyupgrade` should also remove any lines that are empty, or any lines that have an empty comment.

asottile:
it would have to always do that which isn't really in spirit of the tool (the line below is unrelated ). true blank lines are already handled

asottile:
maybe I didn't explain that so great -- `pyupgrade` would need to always do such rewrites to have stable outcomes so independent of the coding pragma it would need to remove that line which would be undesirable

maybe someone likes starting their files with:

```
#
# I am important
#
```

verhovsky:
That change turns

```python
# coding: utf-8
# 
# I am important
# 
```

into

```python
# I am important
# 
```

and

```
# coding: utf-8
#
#
# I am important
# 
```

into 

```python
#
# I am important
# 
```

and so on. You could delete all blank comments in a `for` loop, but that seems a bit much.

```
#
# coding: utf-8
```

stays the same.


```
#
# I am important
#
```

also stays the same.

Maybe I don't understand what you mean by "stable outcomes". I can imagine that deleting an extra line based on the value of a previous line could cause problems, but I can't actually think of a concrete case. 

I guess you mean what if someone starts all of their code with a leading blank comment except for one file where the put a `# coding: utf-8` comment before the leading blank comment. I would guess the people who just put it there because of the coding comment outnumber those people like 10-1, but if you disagree feel free to reject the pull request.

asottile:
yeah I closed this meaning "I don't want this implemented" -- sorry

PR #221, opened 2019-11-01
Delete blank comment line after coding cookie

verhovsky:
https://github.com/asottile/pyupgrade/issues/219

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -746,8 +746,7 @@ def _fix_tokens(contents_text, min_version):
             del tokens[i]
             if tokens[i].name == 'NL':  # pragma: no branch (old PY2)
                 del tokens[i]
-                # pragma: no branch (old PY2)
-                if blank_re.match(tokens[i].src):
+                if blank_re.match(tokens[i].src) and tokens[i].line == 2:
                     del tokens[i]
         elif token.src == 'from' and token.utf8_byte_offset == 0:
             _fix_future_imports(tokens, i, min_version)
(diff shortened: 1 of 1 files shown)


pyupgrade.py
```diff
@@ -745,6 +746,9 @@ def _fix_tokens(contents_text, min_version):
             del tokens[i]
             if tokens[i].name == 'NL':  # pragma: no branch (old PY2)
                 del tokens[i]
+                # pragma: no branch (old PY2)
```
verhovsky: No clue what this line does or if it's necessary.

pyupgrade.py
```diff
@@ -93,8 +93,9 @@ def _convert_tup(tup):  # type: (DotFormatPart) -> str
     return j.join(_convert_tup(tup) for tup in parsed)
 
 
-# if py3+ use tokenize.cookie_re
+# if py3+ use tokenize.cookie_re and tokenize.blank_re
 cookie_re = re.compile('^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)')
+blank_re = re.compile(r'^[ \t\f]*(?:[#\r\n]|$)', re.ASCII)
```
verhovsky: I don't know if the `re.ASCII` part makes sense, I just copy pasted it from `tokenize`.

pyupgrade.py
```diff
@@ -745,6 +746,9 @@ def _fix_tokens(contents_text, min_version):
             del tokens[i]
             if tokens[i].name == 'NL':  # pragma: no branch (old PY2)
                 del tokens[i]
+                # pragma: no branch (old PY2)
+                if blank_re.match(tokens[i].src):
+                    del tokens[i]
```
verhovsky: I think technically there should be another nested check for `if tokens[i].name == 'NL':` here, but I guess the leading empty line trimming takes care of it? Seems to work anyway.

asottile:
via https://github.com/asottile/pyupgrade/issues/219#issuecomment-548774459

PR #223, opened 2019-11-04
Add --version/-V option

atugushev:


asottile:
does this provide anything over `pip freeze | grep pyupgrade`?

atugushev:
Nope, It's the same. I've run `pyupgrade --version` a few times and annoyed that there was no that option. OTOH, it might be too much to bring the new dep for this simple stuff :)

asottile:
:man_shrugging: yeah I think using `pip` is probably sufficient

asottile:
thanks for the PR nonetheless!

PR #234, opened 2020-01-05
README.md: Add usage

cclauss:


asottile:
I'd rather just say "see `--help` for usage" otherwise this will drift.  The individual sections already call out what the options do

asottile:
oh I meant more that I don't think that documenting this in the README is helpful

asottile:
thanks for the patch nonetheless!

PR #260, opened 2020-02-20
Add langage in code sample to have syntax highlighting

laurentbristiel:


asottile:
this was intentionally not done for this block because github considers them (as it should) invalid syntax

asottile:
thanks for the patch nonetheless!

laurentbristiel:
ah ok, thx!

PR #265, opened 2020-02-20
Add some additional modes I missed

asottile:
I noticed I missed a few in #263


chris-rands:
Thanks again, does this correctly handle `gzip.open()` when the default mode is `'rb'`? (worst case, someone did `from gzip import open`)

asottile:
probably not, patches welcome

Issue #249, opened 2020-01-18
Replace open(file, "rU") with open(file)?


chris-rands:
"U" (universal newlines) `mode` was deprecated in Python 3.4 and the `newline=None` argument handles it instead. The default mode is `"r"` anyway

chris-rands:
Thank you!

asottile:
missed some -- added those in #265

tacerus:
> The default mode is "r" anyway

I understand the removal of `U`, but what is the rationale behind removing `r`? Yes, it is not needed, but shouldn't "Explicit is better than implicit." be a valid enough argument to accept it being used?

asottile:
because one way to do things and that's the way I chose

PR #270, opened 2020-03-04
Document overwriting of open modes

verhovsky:
Document changes made for #249


verhovsky:
I don't know if 

> Availability:
> - `--py3-plus` is passed on the commandline.

is correct, I just assumed.

asottile:
yep, that was correct

PR #271, opened 2020-03-06
parens: add support for keeping extraneous parens

durin42:
Mercurial uses extraneous parens as a low-cost way to mark strings as
"intentionally not translated with _()". It's inelegant, but adding
this command-line flag allowed me to run pyupgrade on Mercurial,
fixing a bunch of set/dict comprehensions, set literals, and even
fixing a bug in a regex that was missing an r-prefix on the string.

I'm not sure if I've adequately run your formatting tests etc, because
it wasn't obvious what I should be tweaking. If this functionality isn't
welcome that's fine too, I mostly suspect the use on hg is a
one-and-done for now...

In any case, thanks for pyupgrade!

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -753,7 +753,7 @@ def _fix_import_removals(
                 del tokens[j:idx + 1]
 
 
-def _fix_tokens(contents_text: str, min_version: MinVersion) -> str:
+def _fix_tokens(contents_text: str, min_version: MinVersion, fix_parens: bool) -> str:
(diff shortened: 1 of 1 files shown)


asottile:
hi! glad it helped you and that with your fork you were able to get done what you needed.  I suspect this won't be useful outside of mercurial and it's a bit of work to maintain every new option so I'm not going to accept this at this time (all of the current options are for behaviour-breaks so it doesn't really quite fit in with those).

thanks again for the kind words and the interest in the project though!

Issue #40, opened 2018-06-10
Run on a file or directory


hugovk:
Thanks for this handy tool!

It would be nice to be able to run pyupgrade on a directory, and have it process all .py files in that directory and subdirectories.

For example:
```sh
pyupgrade .
```

This would be similar to other tools like [Flake8](https://github.com/PyCQA/flake8) and [Black](https://github.com/ambv/black/).

In the meantime, this works:

```sh
pyupgrade `find . -name "*.py"`
```

asottile:
I'd suggest using [pre-commit](https://pre-commit.com) to run `pyupgrade` (or, well, any linter) -- pre-commit is better at running your linters than your linters could be because it knows about what is checked in and can identify extensionless executables.

Your "workaround" would be my other suggestion -- no need to make a tool more complicated than it needs to be if there's another tool which does exactly what you need.

Adding directory recursion is a surprising amount of complexity, you then need to handle exclusion, inclusion, globs, walking, ordering, and a bunch of other problems I'm not interested in solving (again) (because `pre-commit` does this better!).

cdeil:
@asottile - thank you for this wonderful tool!

I would also find this option useful to pass a folder. Yes, it's a bit of code to walk the directories, but `black` has a very good implementation that would be copied.

It's not obvious to me why one would use `pre-commit` to upgrade a given codebase. Upgrading is a one-time action, at least that's how I want to use it. So why make this a pre-commit hook? -- @asottile maybe you could clarify a bit at the start of the readme.

hugovk:
My workaround now also checks the found thing is a file (once `*.py` found a directory):

```sh
pyupgrade `find . -name "*.py" -type f`
```

asottile:
It's not just a one time tool, it will continue to enforce the rewrites into the future (and gain features over time). You don't want to have to manually point out "this code base is python 3+ don't do X" in code reviews ad infinitum.

Even if you don't want to have hooks indefinitely, `pre-commit try-repo ... `  is good for one time runs

bagerard:
Hi @asottile I understand that you are advocating for pre-commit but I believe it would be good to mention the trick with `pyupgrade `find . -name "*.py" -type f`` in a "Usage" section of the readme as well. What do you think?

asottile:
[...]

PR #274, opened 2020-03-25
Add more usage instructions to README

adamchainz:
After installing it took me a while to figure out the correct `find` invocation for recursive application to a directory, so I figured that would be useful to document, along with some general instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,31 @@ versions of the language.
 
 `pip install pyupgrade`
 
-## As a pre-commit hook
+## Usage
 
-See [pre-commit](https://github.com/pre-commit/pre-commit) for instructions
+### From the CLI
+
+Run on a single file to upgrade:
+
+```sh
+$ pyupgrade myfile.py
+```
+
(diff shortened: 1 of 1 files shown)


asottile:
hello! thanks for the patch -- however I don't believe it's pyupgrade's readme's responsibility to teach how to use basic unix commands (`--help` / `find` / etc.): https://github.com/asottile/pyupgrade/issues/40#issuecomment-597910147

fwiw, `-exec '{}' +` is a better/faster command than `'{}' ';'` (or to use globs directly and then you don't involve `find` at all!)

adamchainz:
Sure okay, although I added it because many linter programs (flake8, black, isort, etc.) will do directory recursion so I thought it would be handy to have the command ready for users who expect that. I certainly did.

Anyway I opened another PR without #275 since I believe it's worth describing invocation outside of pre-commit.

asottile:
right, but again, `--help` is a standard process to get that information

Issue #280, opened 2020-04-12
Perform rewrite of format literals containing \N escapes


asottile:
#279 prevents the crash / incorrect behaviour from #278, but this should be supported -- shouldn't be too difficult but will likely need to reimplement the format string parsing (the stdlib one doesn't notice these escape sequences, probably because they're gone by that point of the parser)

The current behaviours that should be fixed:

1. % format: `'\N{snowman} %s' % (a,)` should become `'\N{snowman} {}'.format(a)`
2. f-string: `'\N{snowman} {}'.format(a)` should become `f'\N{snowman} {a}'`

peterjc:
To anyone else wondering if the second example is wrong, with a literal ``{snowman}`` inside an f-string, the examples all do this:

```pycon
>>> a = "melts"
>>> '\N{snowman} %s' % (a,)
'☃ melts'
>>> '\N{snowman} {}'.format(a)
'☃ melts'
>>> f'\N{snowman} {a}'
'☃ melts'
```

The escape sequence ``\N{name}`` means the character in the Unicode database https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

But nowadays you could just write ☃ directly in the Python code:

```python
>>> a = "melts"
>>> f'☃ {a}'
```

I guess pyupgrade isn't going to replace escaped emoji with literal emoji, is it?

asottile:
there are situations where that is undesirable:

```pycon
>>> '\N{GREEK CAPITAL LETTER ALPHA}' == 'A'
False
>>> 'Α' == 'A'
False
```

Issue #278, opened 2020-04-12
Valid f-string conversion throws KeyError


gvangool:
While working on a conversion project, I noticed that certain files would throw `KeyError`'s for `\N{...}` strings.

A minimal test case looks like:
```python
world = "world"
print("hello {} \N{snowman}".format(world))
```

Which I would expect to be left alone or rewritten as:
```python
world = "world"
print(f"hello {world} \N{snowman}")
```

I had trouble reproducing it in pytests. This seemed to trigger the same error:
```diff
diff --git tests/fstrings_test.py tests/fstrings_test.py
index d850293..a772a0f 100644
--- tests/fstrings_test.py
+++ tests/fstrings_test.py
@@ -48,6 +48,7 @@ def test_fix_fstrings_noop(s):
         ('"hello {}!".format(name)', 'f"hello {name}!"'),
         ('"{}{{}}{}".format(escaped, y)', 'f"{escaped}{{}}{y}"'),
         ('"{}{b}{}".format(a, c, b=b)', 'f"{a}{b}{c}"'),
+        (r'"{}\N{snowman}".format(a)', r'f"{a}\N{snowman}"'),
         # TODO: poor man's f-strings?
         # '"{foo}".format(**locals())'
     ),
```

asottile:
thanks for the report!  this is indeed broken, I've opened #279 to prevent the crash (plus the incorrect behaviour for %-format rewriting) and #280 to eventually follow up and implement this properly

asottile:
this is "fixed" and released in v2.1.1 -- thanks again for the report!

PR #281, opened 2020-04-12
Rewrite format literals containing \N escapes

gvangool:
This is my attempt at fixing #280 (I started it for #278).

It's based on current parser behaviour:
- for the `%` encoding we let it convert to `\N{{snowman}}` and fix it afterwards
- for the `.format`, the source will end with `\N` (with name being `snowman`). We can reuse the original format for this.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -2422,8 +2422,11 @@ def _fix_py36_plus(contents_text: str) -> str:
             if tokens[end].line != token.line:
                 continue
 
-            tokens[i] = token._replace(src=_to_fstring(token.src, node))
-            del tokens[i + 1:end + 1]
+            try:
+                tokens[i] = token._replace(src=_to_fstring(token.src, node))
+                del tokens[i + 1:end + 1]
+            except KeyError:
+                pass
         elif token.offset in visitor.named_tuples and token.name == 'NAME':
             call = visitor.named_tuples[token.offset]
             types: Dict[str, ast.expr] = {
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -617,8 +617,6 @@ def _fix_format_literal(tokens: List[Token], end: int) -> None:
                     spec is not None and '{' not in spec
             ):
                 last_int += 1
-            elif fmtkey is None and spec is None:
-                pass
             else:
                 return
 
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -64,8 +64,22 @@ def parse_format(s: str) -> Tuple[DotFormatPart, ...]:
     parsed = tuple(_stdlib_parse_format(s))
     if not parsed:
         return ((s, None, None, None),)
-    else:
-        return parsed
+    collapse_part = None
+    new_parsed = []
+    for part in parsed:
+        fragment, fmtkey, _, _ = part
+        if collapse_part:
+            fragment = collapse_part[0] + fragment
+            part = (fragment,) + part[1:]
+            collapse_part = None
+        # rebuild \N escape seqs
+        if ENDS_ON_NAMED_UNICODE_RE.search(fragment) and fmtkey:
+            collapse_part = (fragment + '{' + fmtkey + '}', None, None, None)
+        if collapse_part is None:
+            new_parsed.append(part)
+    if collapse_part:
+        new_parsed.append(collapse_part)
+    return tuple(new_parsed)
 
 
 def unparse_parsed_string(parsed: Sequence[DotFormatPart]) -> str:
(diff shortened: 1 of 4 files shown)


pyupgrade.py
```diff
@@ -917,10 +917,16 @@ def _simplify_conversion_flag(flag: str) -> str:
     return ''.join(parts)
 
 
+NAMED_UNICODE_RE = re.compile(r'\\N{{([^}]+)}}')
```
asottile, line 920: there's a slight issue with this:

```python
a = 1
print(r'\\N{snowman} %s' % (a,))
```

```console
$ python3 t.py.old
\\N{snowman} 1
```

```console
$ cp t.py.old t.py && pyupgrade t.py
Rewriting t.py
$ python3 t.py
  File "t.py", line 2, in <module>
    print(r'\\N{snowman} {}'.format(a))
KeyError: 'snowman'
```

I believe you need something like this: https://github.com/asottile-archive/textmate-highlight/commit/c52ec58b7cdde404baa0e53cf1b26a940bbcc4e4#diff-bcc683052c071bec26ad5c7ed6fc5d1fR44

pyupgrade.py
```diff
@@ -2229,7 +2231,7 @@ def _to_fstring(src: str, call: ast.Call) -> str:
     parts = []
     i = 0
     for s, name, spec, conv in parse_format('f' + src):
-        if name is not None:
+        if name is not None and not s.endswith(r'\N'):  # skip unicodedata.lookup
```
asottile, line 2234: this might have the same problem as above, not sure though 🤔

gvangool: It doesn't, since `r'"{} \\N{snowman}".format(a)'` is an invalid format string.

I've added it to the noop tests though.

pyupgrade.py
gvangool: Thanks for the pointer, negative lookbehind (`(?<!...)`) was exactly what I needed.

pyupgrade.py
asottile: what about `r'{foo} \N{snowman}'.format(foo=a, snowman=b)`

tests/fstrings_test.py
```diff
@@ -30,6 +30,7 @@
         '"{}{}".format(a)', '"{a}{b}".format(a=a)',
         # invald \N escape
         r'"{} \\N{snowman}".format(a)',
+        r'"{} \\N{snowman}".format(a, snowman)',
```
asottile, line 33: can we show the successful case as well, where it gets turned into an f-string but slightly looks like a \N escape?

pyupgrade.py
```diff
@@ -2225,13 +2225,18 @@ def _unparse(node: ast.expr) -> str:
         raise NotImplementedError(ast.dump(node))
 
 
+ENDS_ON_NAMED_UNICODE_RE = re.compile(r'(?<!\\)\\N$')
```
asottile, line 2228: hmm, this regex doesn't match `\\\N`

pyupgrade.py
```diff
@@ -917,10 +917,16 @@ def _simplify_conversion_flag(flag: str) -> str:
     return ''.join(parts)
 
 
+NAMED_UNICODE_RE = re.compile(r'(?<!\\)\\N{{([^}]+)}}')
```
asottile, line 920: this one also doesn't work, I was hoping you'd copy the example I linked :) -- it needs to match (1) negative lookbehind for backslash (2) even-number-of-backslashes (3) \N

asottile:
looks good, looks like there's one last thing to fix 🤔 

```diff
diff --git a/tests/format_literals_test.py b/tests/format_literals_test.py
index f3610a1..a809f74 100644
--- a/tests/format_literals_test.py
+++ b/tests/format_literals_test.py
@@ -49,7 +49,7 @@ def test_intentionally_not_round_trip(s, expected):
         "'{' '0}'.format(1)",
         # comment looks like placeholder but is not!
         '("{0}" # {1}\n"{2}").format(1, 2, 3)',
-        # TODO: this works by accident (extended escape treated as placeholder)
+        # this works by accident (extended escape treated as placeholder)
         r'"\N{snowman} {}".format(1)',
     ),
 )
@@ -99,6 +99,8 @@ def test_format_literals_noop(s):
         ),
         # parenthesized string literals
         ('("{0}").format(1)', '("{}").format(1)'),
+        # extended unicode escapes
+        (r'"\N{snowman} {0}".format(1)', r'"\N{snowman} {}".format(1)'),
     ),
 )
 def test_format_literals(s, expected):
```

pyupgrade.py
```diff
@@ -586,15 +589,18 @@ def _fix_format_literal(tokens: List[Token], end: int) -> None:
...
             return
 
         # The last segment will always be the end of the string and not a
         # format, slice avoids the `None` format key
-        for _, fmtkey, spec, _ in parsed[:-1]:
-            if (
+        for i in range(len(parsed) - 1):
+            s, fmtkey, spec, _ = parsed[i]
+            if ENDS_ON_NAMED_UNICODE_RE.search(s):  # rebuild \N escape seqs
+                parsed[i] = (s + '{' + fmtkey + '}', None, None, None)
```
asottile, line 602: this feels like a hack :/ -- I'd prefer if this lived in `parse_format` itself (would probably simplify some of the other cases as well?)

pyupgrade.py
```diff
@@ -586,15 +589,18 @@ def _fix_format_literal(tokens: List[Token], end: int) -> None:
...
     for i in parts:
         try:
-            parsed = parse_format(tokens[i].src)
+            parsed = list(parse_format(tokens[i].src))
         except ValueError:
             # the format literal was malformed, skip it
             return
 
         # The last segment will always be the end of the string and not a
         # format, slice avoids the `None` format key
```
asottile, line 598: this comment is orphaned

pyupgrade.py
```diff
@@ -2429,8 +2447,11 @@ def _fix_py36_plus(contents_text: str) -> str:
             if tokens[end].line != token.line:
                 continue
 
-            tokens[i] = token._replace(src=_to_fstring(token.src, node))
-            del tokens[i + 1:end + 1]
+            try:
+                tokens[i] = token._replace(src=_to_fstring(token.src, node))
+                del tokens[i + 1:end + 1]
+            except KeyError:
+                pass
```
gvangool, line 2454: I was looking for an unparse-like function that could be used in the `_to_fstring` because this feels like it's handling an implementation detail.

PR #294, opened 2020-04-26
Remove erroneous string in test cases

mxr:


asottile:
this is intentional, a noop test case that the empty string is not changed

mxr:
TIL

mxr:
Why is it in that set of test cases?

asottile:
the noop tests?

mxr:
It's in the named tuple tests file, but `''` isn't specific to named tuples?

asottile:
it broke while I was implementing the namedtuple check

mxr:
🤯

PR #295, opened 2020-04-27
Rewrite mock to unittest.mock

### LINKED ISSUE #233: Replace `from mock import` with `from unittest.mock import`
https://docs.python.org/library/unittest.mock.html it was added in 3.3 as `unittest.mock` and backported to 2 as `mock`. Some codebases that used to be Python 2, still have `pip install mock` and `import mock` or `from mock import Mock` when they could forgo the install and do `from unittest.mock import Mock`.

mxr:
Resolves #233

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1189,6 +1189,8 @@ class FindPy3Plus(ast.NodeVisitor):
 
     FROM_IMPORTED_MODULES = OS_ERROR_ALIAS_MODULES.union(('functools', 'six'))
 
+    MOCK_MODULES = frozenset(('mock', 'mock.mock'))
+
     class ClassInfo:
         def __init__(self, name: str) -> None:
             self.name = name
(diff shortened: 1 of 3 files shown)


pyupgrade.py
```diff
@@ -1210,6 +1217,7 @@ def __init__(self) -> None:
 
         self._from_imports: Dict[str, Set[str]] = collections.defaultdict(set)
         self.io_open_calls: Set[Offset] = set()
+        self.mock_simple: Set[Offset] = set()
```
asottile, line 1220: mock_imports maybe?

pyupgrade.py
```diff
@@ -1359,6 +1384,8 @@ def _visit_comp(self, node: ast.expr) -> None:
     def _visit_simple(self, node: NameOrAttr) -> None:
         if self._is_six(node, SIX_SIMPLE_ATTRS):
             self.six_simple[_ast_to_offset(node)] = node
+        elif self._is_mock(node, self.MOCK_ALL):
+            self.mock_simple.add(_ast_to_offset(node))
```
asottile, line 1388: shouldn't this just replace `import mock` with `from unittest import mock`?

mxr: Sure

pyupgrade.py
```diff
@@ -1289,13 +1291,23 @@ def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
             for name in node.names:
                 if not name.asname:
                     self._from_imports[node.module].add(name.name)
+        elif node.module == 'mock':
```
asottile, line 1294: could also do `mock.mock` here right?

mxr: Good point

PR #296, opened 2020-04-28
Rewrite @functools.lru_cache()

### LINKED ISSUE #229: Remove parentheses from @functools.lru_cache() when upgrading to Python 3.8
https://docs.python.org/library/functools.html#functools.lru_cache

In Python 3.7 you need to write `@lru_cache()`, in 3.8 you can just do `@lru_cache` without the parentheses. See https://bugs.python.org/issue36772

pyupgrade could remove the unnecessary parens for this and other decorators in the stdlib. 

If someone monkey-patches `lru_cache` in their code with an implementation that doesn't handle not having parens this would break their code.

mxr:
Resolves #229

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1187,7 +1187,7 @@ class FindPy3Plus(ast.NodeVisitor):
         'socket',
     ))
 
-    FROM_IMPORTED_MODULES = OS_ERROR_ALIAS_MODULES.union(('six',))
+    FROM_IMPORTED_MODULES = OS_ERROR_ALIAS_MODULES.union(('functools', 'six'))
 
     class ClassInfo:
         def __init__(self, name: str) -> None:
@@ -1233,6 +1233,8 @@ class FindPy3Plus(ast.NodeVisitor):
         self._in_async_def = False
         self.yield_from_fors: Set[Offset] = set()
 
+        self.no_arg_decorators: Set[Offset] = set()
+
     def _is_six(self, node: ast.expr, names: Container[str]) -> bool:
         return (
             isinstance(node, ast.Name) and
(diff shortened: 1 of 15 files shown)


pyupgrade.py
```diff
@@ -2480,6 +2543,8 @@ def _fix_file(filename: str, args: argparse.Namespace) -> int:
         contents_text = _fix_py3_plus(contents_text)
     if args.min_version >= (3, 6):
         contents_text = _fix_py36_plus(contents_text)
+    if args.min_version >= (3, 8):
+        contents_text = _fix_py36_plus(contents_text)
```
asottile, line 2547: each parse / unparse adds a few 100ms to rewriting -- ideally we'd combine these in existing classes

mxr: Sure

pyupgrade.py
```diff
@@ -2477,7 +2505,7 @@ def _fix_file(filename: str, args: argparse.Namespace) -> int:
     if not args.keep_percent_format:
         contents_text = _fix_percent_format(contents_text)
     if args.min_version >= (3,):
-        contents_text = _fix_py3_plus(contents_text)
+        contents_text = _fix_py3_plus(contents_text, args.min_version)
```
asottile, line 2508: this one might be simple enough to do with `_fix_tokens` -- then you won't need to rewrite every call to `_fix_py3_plus`

mxr: If I do that would I just search for `lru_cache()`? If so, how would I avoid rewriting `from not_functools import lru_cache; @lru_cache()`?

mxr: I could make `min_version` an optional arg to avoid rewriting `_fix_py3_plus` calls

mxr:
Is it worth making `min_version` an optional arg as a followup?

asottile:
no, the only callers are internal

PR #302, opened 2020-05-05
Rewrite @dataclasses.dataclass()

mxr:


asottile:
hmmm weren't dataclasses introduced in python3.7?

mxr:
I thought that the rewrite is valid in at least 3.7.0?

```console
$ python --version
Python 3.7.0
$ python -c """
dquote> from dataclasses import dataclass

@dataclass
class Data:
    foo: int
    bar: str

print(Data(1, '2'))
"""
Data(foo=1, bar='2')
```

asottile:
3.7.0 was when dataclasses were introduced, so there's nothing to "upgrade"

PR #304, opened 2020-05-11
Run pre-commit autoupdate

asottile:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,25 +11,25 @@ repos:
     -   id: requirements-txt-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/asottile/setup-cfg-fmt
-    rev: v1.6.0
+    rev: v1.9.0
     hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -11,25 +11,25 @@ repos:
...
-    rev: v1.5
+    rev: v1.5.2
     hooks:
     -   id: autopep8
 -   repo: https://github.com/asottile/reorder_python_imports
-    rev: v1.9.0
+    rev: v2.3.0
     hooks:
     -   id: reorder-python-imports
         args: [--py3-plus]
```
mxr, line 30:
```suggestion
        args: [--py36-plus]
```

asottile: yeah I'll do that in another pass, though it doesn't change things here

need to track down the regression in python3.9 though >.<

asottile:
blocked on https://github.com/nedbat/coveragepy/issues/988

PR #320, opened 2020-06-19
Rewrite old-style-class super calls

asottile:


ThiefMaster:
I think this kind of rewriting may be dangerous or at least not what people expect (so being able to opt-in/out from it would probably be good).

For example, in my application we make heavy use of method calls of parent classes access checks. Usually we call the methods of all parent classes (sometimes in a specific order to being very explicit is beneficial), sometimes only one of them (because we explicitly do not want the access check from the other parent), and sometimes there is only one parent class.

And the code here looks like my last case would always be rewritten, but that would make the code base inconsistent depending on how many parent classes there are - and I believe some place in the Python docs even tells you to avoid mixing super and classic parent method calling within a single inheritance chain.

asottile:
@ThiefMaster your assumption is incorrect, this rewrite will only happen when:
- the class has a single unambigous base
- the call is not within another scope (lambda, nested function, comprehension)
- the function name exactly matches the current function

in this case the call to super() is identical and safe.  there will not be an option for this -- if you don't want this then fork the tool

PR #326, opened 2020-07-04
Include tests in sdist

sebix:
Includes the tests in the source tarball, which is mostly for dev and packaging purpose. Tests are needed/useful for further packaging.

The tests will not be included in the bdist_wheel packages.


asottile:
no thanks, sorry -- use the github distribution if you need them

PR #335, opened 2020-08-08
Add function to fix str in memory.

Carreau:
This make it slighlty more convenient to fix code snippets in memory and
for example run pyupgrade, black and isort at once without writing to
disk.


asottile:
I don't want to support this sorry, the only api is the command line

Carreau:
No problem. Thanks.

Issue #339, opened 2020-09-18
Recursive pyupgrade option


dsaxton:
Very cool project @asottile! I was curious if there was a way to run pyupgrade on whole directories or projects, and if not whether this might be a helpful feature to add? For instance have `pyupgrade -r some_directory` run recursively on the directory, or even drop the `-r` and let the recursive part be implied?

Thanks for building this!

asottile:
dupe #150 #46

asottile:
`pre-commit run pyupgrade --all-files`

saroad2:
> `pre-commit run pyupgrade --all-files` or shell globs `pyupgrade **/*.py`

I don't think that this is sufficient solution, because now you are forcing the user to use `pre-commit` instead of simply using Pyupgrade as a standalone tool.

For instance, I would love to add Pyupgrade as one of the default tools run by [Statue](https://github.com/saroad2/statue), but the absence of the ability to run on directories is the only thing preventing me from doing it.

As I've shown in #344 , Implementation of this ability is quite easy and I hope you reconsider.

Issue #150, opened 2019-05-22
Support use on directories


Zac-HD:
In all my uses of `pyupgrade`, I want to apply it to every Python file in the repo - in general, some directory.  `pyupgrade $(find . -name '*.py')` works OK on \*nix systems, but I'd be happier if I could just write `pyupgrade .` and have it work on Windows too!

Personally I'd prefer to follow Black and flake8 (etc) by having recursive-by-default for directories, but if not a `-r`/`--recursive` flag is also fairly common.

asottile:
see #40 #46 https://github.com/asottile/add-trailing-comma/pull/51 https://github.com/asottile/add-trailing-comma/issues/59 https://github.com/asottile/reorder_python_imports/pull/76

Issue #46, opened 2018-07-15
Recursive mode?


alok:
Currently I run pyupgrade with something like `find ... | xargs pyupgrade`. A recursive flag to drill down directories would be nice (and in line with a lot of other formatters like black, yapf, and isort).

alok:
Oh that's neat! I didn't know about the `--all-files` option. That's perfect.

impredicative:
To offer some contrast, I think using `pre-commit` opens a whole can of worms, and it's shameless to need to require it. The following is simple and works in Bash:

```bash
shopt -s globstar
pyupgrade ./**/*.py
```

PR #344, opened 2020-10-16
Pyupgrade can now support directories as input

saroad2:
Pyupgrade should be able to handle directories as inputs, like any other static code analysis tools (pylint, flake8, mypy, black, etc.)

Added this ability 😃.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -4,11 +4,13 @@ import codecs
 import collections
 import contextlib
 import keyword
+import os
 import re
 import string
 import sys
 import tokenize
 import warnings
+from pathlib import Path
 from typing import Any
 from typing import cast
 from typing import Container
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
This looks like a duplicate of #150 #46 #339

saroad2:
Hey @MarcoGorelli ,
I figured out that even though a lot of people have opened issues about recursive Pyupgrade, nobody actually did a PR to implement it. I thought that if I "just do it", than maybe the maintainers of Pyupgrade will consider merging it.
It does not damage any of the current functionalities, just adding a new one.

Moreover, I should state that the only reason I don't integrate Pyupgrade with my own projects, especially [Statue](https://github.com/saroad2/statue), is because of the absence of this ability. I hope that now that I've shown that this is possible, you'd reconsider.

PR #350, opened 2020-10-28
fixed #349 - parenthesis for BoolOps

### LINKED ISSUE #349: Six call removal breaks logic inside call
before:
```
import six                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                               
result = {}                                                                                                                                                                                                                                    
for leg in six.itervalues(result.get('something') or {}):                                                                                                                                                                                      
    pass
```
after:
```
import six                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                               
result = {}                                                                                                                                                                                                                                    
for leg in result.get('something') or {}.values():                                                                                                                                                                                             
    pass 
```
should:
```
import six                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                               
result = {}                                                                                                                                                                                                                                    
for leg in (result.get('something') or {}).values():                                                                                                                                                                                             
    pass

### PR DESCRIPTION
**Pull request by tobixx**
Implemented as you wished, but the `needs_parenthesis` check right now only looks for `BoolOp`, `expr` wasn't the right thing.

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1536,12 +1536,12 @@ class FindPy3Plus(ast.NodeVisitor):
         elif self._is_six(node.func, ('b', 'ensure_binary')):
             self.six_b.add(_ast_to_offset(node))
         elif self._is_six(node.func, SIX_CALLS) and not _starargs(node):
-            needs_parenthesis = node.args and isinstance(
+            offset = _ast_to_offset(node)
+
+            if node.args and isinstance(
                 node.args[0],
                 (ast.BoolOp, ast.IfExp, ast.BinOp, ast.GeneratorExp),
-            )
-            offset = _ast_to_offset(node)
-            if needs_parenthesis:
+            ):
                 self.six_calls_parenthesized[offset] = node
             else:
                 self.six_calls[offset] = node
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -1122,6 +1122,9 @@ SIX_CALLS = {
     'assertRaisesRegex': '{args[0]}.assertRaisesRegex({rest})',
     'assertRegex': '{args[0]}.assertRegex({rest})',
 }
+SIX_CALLS_PARENTHESIZED = {
+    k: v.replace('{args[0]}', '({args[0]})') for k, v in SIX_CALLS.items()
+}
 SIX_B_TMPL = 'b{args[0]}'
 WITH_METACLASS_NO_BASES_TMPL = 'metaclass={args[0]}'
 WITH_METACLASS_BASES_TMPL = '{rest}, metaclass={args[0]}'
@@ -1244,6 +1247,7 @@ class FindPy3Plus(ast.NodeVisitor):
         self.six_add_metaclass: Set[Offset] = set()
         self.six_b: Set[Offset] = set()
         self.six_calls: Dict[Offset, ast.Call] = {}
+        self.six_calls_parenthesized: Dict[Offset, ast.Call] = {}
         self.six_iter: Dict[Offset, ast.Call] = {}
         self._previous_node: Optional[ast.AST] = None
         self.six_raise_from: Set[Offset] = set()
@@ -1535,7 +1539,15 @@ class FindPy3Plus(ast.NodeVisitor):
         elif self._is_six(node.func, ('b', 'ensure_binary')):
             self.six_b.add(_ast_to_offset(node))
         elif self._is_six(node.func, SIX_CALLS) and not _starargs(node):
-            self.six_calls[_ast_to_offset(node)] = node
+            needs_parenthesis = (
+                node.args and
+                isinstance(node.args[0], ast.BoolOp)
+            )
+            six_calls = (
+                self.six_calls_parenthesized
+                if needs_parenthesis else self.six_calls
+            )
+            six_calls[_ast_to_offset(node)] = node
         elif (
                 isinstance(node.func, ast.Name) and
                 node.func.id == 'next' and
@@ -2062,6 +2074,7 @@ def _fix_py3_plus(
             visitor.six_add_metaclass,
             visitor.six_b,
             visitor.six_calls,
+            visitor.six_calls_parenthesized,
             visitor.six_iter,
             visitor.six_raise_from,
             visitor.six_reraise,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pyupgrade.py [hunk 1]
```diff
@@ -1122,6 +1122,9 @@ def _fix_percent_format(contents_text: str) -> str:
     'assertRaisesRegex': '{args[0]}.assertRaisesRegex({rest})',
     'assertRegex': '{args[0]}.assertRegex({rest})',
 }
+SIX_CALLS_PARENTHESIZED = {
+    k: v.replace('{args[0]}', '({args[0]})') for k, v in SIX_CALLS.items()
+}
```
asottile, line 1127: this isn't quite correct, some of the calls in `SIX_CALLS` do not need this treatment (for example `int2byte`)

pyupgrade.py
```diff
@@ -1535,7 +1539,15 @@ def visit_Call(self, node: ast.Call) -> None:
...
-            self.six_calls[_ast_to_offset(node)] = node
+            needs_parenthesis = (
+                node.args and
+                isinstance(node.args[0], ast.BoolOp)
+            )
+            six_calls = (
+                self.six_calls_parenthesized
+                if needs_parenthesis else self.six_calls
+            )
+            six_calls[_ast_to_offset(node)] = node
```
asottile, line 1550: please write an if statement instead of this indirection -- it'll be much more readable

pyupgrade.py
```diff
@@ -2161,12 +2174,21 @@ def _replace(i: int, mapping: Dict[str, str], node: NameOrAttr) -> None:
             assert isinstance(call.func, (ast.Name, ast.Attribute))
             template = f'iter({_get_tmpl(SIX_CALLS, call.func)})'
             _replace_call(tokens, i, end, func_args, template)
-        elif token.offset in visitor.six_calls:
+        elif (
+            token.offset in visitor.six_calls or
+            token.offset in visitor.six_calls_parenthesized
+        ):
```
asottile, line 2180: I think this should be two separate if statements

pyupgrade.py
```diff
@@ -1535,7 +1539,15 @@ def visit_Call(self, node: ast.Call) -> None:
         elif self._is_six(node.func, ('b', 'ensure_binary')):
             self.six_b.add(_ast_to_offset(node))
         elif self._is_six(node.func, SIX_CALLS) and not _starargs(node):
-            self.six_calls[_ast_to_offset(node)] = node
+            needs_parenthesis = (
+                node.args and
+                isinstance(node.args[0], ast.BoolOp)
```
asottile, line 1544: BinOp, IfExp, GeneratorExp is also susceptible to this

asottile (commented):
looks good for a start, I'd move the code a little bit and it looks like there's a few more cases to handle

pyupgrade.py
tobixx: Well, of course. My arguments against it were code duplication, no effect where it does not apply, prevention of missing entries, lower maintenance. Due to the stable state of `six` in reality I will not fight for it now, but I will come back on this once Python4 comes around and `six`will experience a revival ;-)

asottile: you don't need any duplicate information, the parenthesizing can be done as an option to the formatter

tobixx: Uh - and also I have no clue for which one it applies really (beside my specific bug report and your mentioned one) ...

asottile: ah ok, it's pretty easy to see from the mapping there -- imagine putting `1 + 2` for `args[0]` and see which ones would be syntactically different than `(1 + 2)` for `args[0]` and then you know which ones to specialize

tests/six_test.py
```diff
@@ -382,6 +382,26 @@ def test_fix_six_noop(s):
...
+            '({} if True else {}).keys()\n',
+            id='six.itervalues with if-expression inside call',
+        ),
+        pytest.param(
+            'six.iteritems({} for _ in "")\n',
+            '({} for _ in "").items()\n',
+            id='six.itervalues with generator expression inside call',
+        ),
+        pytest.param(
+            'six.int2byte(1 & 1)\n',
```
asottile, line 401: int2byte is one of the few that *doesn't* need parens added

pyupgrade.py
```diff
@@ -2029,6 +2038,19 @@ def _replace_yield(tokens: List[Token], i: int) -> None:
     tokens[i:block.end] = [Token('CODE', f'yield from {container}\n')]
 
 
+def _replace_six_calls(
+    tokens: List[Token], call: ast.Call, i: int,
+    needs_parenthesis: bool = False,
```
asottile, line 2043: let's make this keyword-only so it isn't a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

pyupgrade.py
```diff
@@ -1535,7 +1536,15 @@ def visit_Call(self, node: ast.Call) -> None:
         elif self._is_six(node.func, ('b', 'ensure_binary')):
             self.six_b.add(_ast_to_offset(node))
         elif self._is_six(node.func, SIX_CALLS) and not _starargs(node):
-            self.six_calls[_ast_to_offset(node)] = node
+            needs_parenthesis = node.args and isinstance(
+                node.args[0],
+                (ast.BoolOp, ast.IfExp, ast.BinOp, ast.GeneratorExp),
+            )
```
asottile, line 1542: this is only used once, I'd inline it in the `if`:

```python
    if (
            node.args and
            isinstance(node.args[0], (...)
    ):
```

pyupgrade.py
tobixx: Uhm, are you sure ? Needs default values (None) -> needs `Optional` typing, needs None checks ... looks, well, non-optimal

tests/six_test.py
tobixx: Well, I just added this test for your mentioned `BinOp` - because my imagination is not big enough to come up wit a different one, beside a 3.9 only dict merge.

pyupgrade.py
```diff
@@ -1535,7 +1536,15 @@ def visit_Call(self, node: ast.Call) -> None:
         elif self._is_six(node.func, ('b', 'ensure_binary')):
             self.six_b.add(_ast_to_offset(node))
         elif self._is_six(node.func, SIX_CALLS) and not _starargs(node):
-            self.six_calls[_ast_to_offset(node)] = node
+            offset = _ast_to_offset(node)
+
+            if node.args and isinstance(
+                node.args[0],
+                (ast.BoolOp, ast.IfExp, ast.BinOp, ast.GeneratorExp),
+            ):
```
asottile, line 1544: I spent time to write the code for you in a comment with my preferred style, why did you pick a different style :S

pyupgrade.py
asottile: yes I'm sure, but you're not understanding me correctly -- check out this video https://youtu.be/aKCfCmSggPg?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY

tests/six_test.py
asottile: that's fine, the other examples aren't valid calls either (itervalues on a generator for example) -- feel free to use a merge operator

but also keep this test and fix the output since it's adding an extra set of parens incorrectly currently

pyupgrade.py
tobixx: Oh - that one, got it.

pyupgrade.py
tobixx: Well, I really thought a lot of how to answer this statement (which already may point into the direction of the problem with it), but I will throw them all away and just answer it with some technical reason: `black` overruled you.

tests/six_test.py
tobixx: Done. But regarding the generator: I only came up with one idea - which is valid in Python3 but not in six for Python2. Any other suggestion?

asottile:
sorry, I have no intention of accepting a patch if you're going to be disrespectful

take your attitude somewhere else

Issue #356, opened 2020-11-09
Provide better options to integrate with CI


ThiefMaster:
There's a non-zero exit code if something was changed (according to https://github.com/asottile/pyupgrade/issues/310#issuecomment-630282643), but when running multiple tools (isort, flake8, etc.) then getting files modified in-place is a huge mess because any command that runs later will then act on already-modified code.

Having a `--check` option that prevents files from being written and maybe even a `--diff` option that prints a diff of what would be changed would help with this a lot.

I wouldn't mind sending a PR for this, but before spending time on this I'd like to know if this will actually get accepted. :)

In case you wonder why I want to run this in CI: A pre-commit hook is great, but only when people use it. Sometimes you get PRs from people who don't bother setting up a proper development environment so they never run the pre-commit hooks. So I'd like CI to fail in such cases.

asottile:
use `pre-commit run --all-files --show-diff-on-failure`

please search the issue tracker next time: #310 #210 (and many more)

Issue #310, opened 2020-05-18
Add --check flag


saroad2:
I would like to use this tool not only as a formatting tool, but also as a checker in my CI/CD process, making sure that people use the latest APIs in their code.

In order to achieve that I would like to add a `--check` flag that cause the following behavior:
1. If no changes are needed in the given sources, return 0 exit code
2. If changes are needed **do not make those changes**, but only return error exit code.

hugovk:
It would be useful but I expect this will be closed as a duplicate of https://github.com/asottile/pyupgrade/issues/210, so here's a workaround:

```sh
pyupgrade *.py --py36-plus  # or however you run it
test $(git status --porcelain | wc -l) = 0
```

If changes are needed, it _will_ actually make those changes (shouldn't matter too much on CI). And assuming your code is in Git, and nothing else has modified it, the second line will return 1 and fail the CI if changes were made. Otherwise it returns 0.

---

Alternatively, use it as part of pre-commit with something like this in `.pre-commit-config.yaml`:

```yml
repos:
  - repo: https://github.com/asottile/pyupgrade
    rev: v2.4.3
    hooks:
      - id: pyupgrade
        args: ["--py36-plus"]
```

And then run this on the CI:

```sh
pip install pre-commit
pre-commit run all-files
```

asottile:
pyupgrade also exits nonzero when making changes so the second line is unnecessary

Issue #258, opened 2020-02-19
Feature request: dry-run option


SWu:
Would it be possible to add an option that doesn't modify any files in-place, but instead just prints a diff of what would have changed and exits 1?

asottile:
Duplicate #210

abompard:
I would like to ask you to reconsider accepting this feature. I'm interested in bringing pyupgrade to my team's tox checks, but bringing in a new tool that changes the code they write is too big of a step. It would make adoption easier for me if I could add a tox test that just reports on files needing upgrades, at least for now.

asottile:
you can always `git checkout -- .`

Issue #210, opened 2019-10-16
Add --check option


decaz:
It would be nice to have such an option when in place changes are unwanted.

asottile:
(I can't find the dupe at the moment, but there is one somewhere) -- sorry I'm not interested in supporting this option -- thanks for the issue nonetheless!

impredicative:
@asottile I'm seeing a lot of lot of reasonable user requests in this repo that you've closed for no good reason other than they don't suit your whim. Why did you even make this project open source? A `--check` flag would obviously be useful for static analysis purposes as a part of a testing script.

asottile:
> I'm seeing a lot of lot of reasonable user requests in this repo that you've closed for no good reason other than they don't suit your whim. Why did you even make this project open source? 

first off, happy holidays -- glad you found the project useful.

please be respectful

PR #370, opened 2021-01-06
Add `--check` option

lummax:
This adds a `pyupgrade --check` option that does only check but not
rewrite the given files. If a file would be rewritten, a message is
printed to `stderr`.

Alternatives
------------

1. Using [pre-commit](https://pre-commit.com/)

Instead of adding this option, the user could use
`pre-commit run --all-files --show-diff-on-failure`.

2. Querying the version control system

Instead of adding this option, the user could let `pyupgrade` change the
files and use `git` or a similar tool to determine if something has
changed. After that the version control system could revert the changes.

3. Writing an orchestration script

Instead of adding this option, the user could write their own
orchestration script that would open and read the file to be checked and
pass them via `stdin` to `pyupgrade`. It could then compare the output
to the original content.

Why this is useful
------------------

A `--check` option is useful whenever one wants to enforce that all code
is compliant to a certain standard. This applies to `git` `pre-commit`
hooks, but also automated tests (especially run in a CI context).

Adding this option provides a significant benefit while adding minimal
additional maintenaince burden. The alternatives mentioned above all
have their own shortcomings:

1. In certain restricted environments it is not possible to load
arbitrary code during a CI run. Especially not from untrusted sources
(`pypi`, `npm` or even github). Additionally in some environments the
user might want to lock down exact versions of tools to be used. While
possible with `pre-commit`, it requires quite some overhead to install
and configure `pre-commit` and create and maintain a local
hook-repository just to be able to run `pyupgrade` as part of an
automated CI test.

2. This is simply not a good idea when it comes to parallel test
execution when we imagine that multiple tools modify files.

3. This is an actual alternative with two problems: It requires every
user to write this orchestration themselves while providing a `--check`
option is minimally intrusive. It requires a `pyupgrade` invocation for
every file to be checked. This can become quite costly in environments
with many python files.

Alternative implementation
--------------------------

Instead of just printing the filename that would be rewritten, the tool
could also print a diff that informs the user about the changes to be
made. This idea was discarded as it introduces a higher maintenaince
burden.

```python
import difflib

print(
    '\n'.join(
        difflib.unified_diff(
            contents_text_orig.splitlines(),
            contents_text.splitlines(),
            fromfile=f'a/{filename}',
            tofile=f'b/{filename}',
        )
    )
)
```

Related
-------

- Add --check option #210
- Feature request: dry-run option #258
- Add --check flag #310
- Provide better options to integrate with CI #356

### DIFF
--- a/pyupgrade.py
+++ b/pyupgrade.py
@@ -2717,7 +2717,9 @@ def _fix_file(filename: str, args: argparse.Namespace) -> int:
     if args.min_version >= (3, 6):
         contents_text = _fix_py36_plus(contents_text)
 
-    if filename == '-':
+    if args.check:
+        print(f'Would rewrite {filename}', file=sys.stderr)
+    elif filename == '-':
         print(contents_text, end='')
     elif contents_text != contents_text_orig:
         print(f'Rewriting {filename}', file=sys.stderr)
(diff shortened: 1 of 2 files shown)


asottile:
did you read the issues? no

Issue #343, opened 2020-09-28
Replace typing.List with list and others


verhovsky:
https://docs.python.org/3.9/whatsnew/3.9.html#pep-585-builtin-generic-types

Python 3.9 (and Python 3.7 if you do `from __future__ import annotations`) will let you write

```
from typing import List

def a(my_arg: List) -> List:
    return []
```

as

```python
def a(my_arg: list) -> list:
    return []
```

Same goes for `typing.Dict` and a bunch of others, see https://www.python.org/dev/peps/pep-0585/

graingert:
it might be worth expanding this issue into a general modern `typing` syntax epic.

upgrading conditional typing imports for new python version, or do people want it conditional still because `typing` can be slow to import?

should pyupgrade be upgrading redundant forward-references, in presence of the `from __future__ import annotations`, redundant forward-references seems like it might be a [pylint's](https://github.com/PyCQA/pylint/issues/3874) or pyflakes responsibility

what about type comments into full fledged annotations: `v: # type: ...` into `v: ...`?
and how about use of third party typing shims: `v: attr.ib(type=...)` into  `v: ... = attr.ib()`?

verhovsky:
> type comments into full fledged annotations

https://github.com/ilevkivskyi/com2ann/blob/master/src/com2ann.py does this.

verhovsky:
This shouldn't be implemented until mypy supports it https://github.com/python/mypy/issues/7907

asottile:
agreed 👍

jdufresne:
Perhaps another example of the same idea, replace `typing.Text` with `str` for `--py3-plus`.

Example input:

```py
if False:
    from typing import Text


def f():
    # type: () -> Text
    return "hello"
```

Example diff:

```diff
--- before.py
+++ after.py
@@ -3,5 +3,5 @@
 
 
 def f():
-    # type: () -> Text
+    # type: () -> str
     return "hello"

```

asottile:
via #371

twoertwein:
> Same goes for `typing.Dict` and a bunch of others, see https://www.python.org/dev/peps/pep-0585/

Are there plans to also re-write `Dict`, `Tuple`, ... for python 3.9+?

asottile:
@twoertwein it's already implemented

twoertwein:
Thanks, I just noticed that :)

[...]

PR #371, opened 2021-01-15
Add --py39-plus and pep585 rewriting

asottile:
See #343

this is just the start of the rewriting here, still need to do a lot with import renaming


graingert:
Thanks so much, I was dreading having to do this by hand!

PR #383, opened 2021-01-30
replace typing.Text with str

### LINKED ISSUE #374: replace `typing.Text` with `str` for `--py3-plus`
Perhaps another example of the same idea, replace `typing.Text` with `str` for `--py3-plus`.

Example input:

```py
if False:
    from typing import Text


def f():
    # type: () -> Text
    return "hello"
```

Example diff:

```diff
--- before.py
+++ after.py
@@ -3,5 +3,5 @@
 
 
 def f():
-    # type: () -> Text
+    # type: () -> str
     return "hello"

```

_Originally posted by @jdufresne in https://github.com/asottile/pyupgrade/issues/343#issuecomment-750930060_

MarcoGorelli:
closes #374

### DIFF
--- a/pyupgrade/_plugins/typing_text.py
+++ b/pyupgrade/_plugins/typing_text.py
@@ -1,0 +1,51 @@
+import ast
+import functools
+from typing import Iterable
+from typing import Tuple
+
+from tokenize_rt import Offset
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import replace_name
+
+
+@register(ast.Attribute)
+def visit_Attribute(
+        state: State,
+        node: ast.Attribute,
+        parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if (
+            state.settings.min_version >= (3,) and
+            isinstance(node.value, ast.Name) and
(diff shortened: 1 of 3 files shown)


pyupgrade/_plugins/typing_text.py
```diff
@@ -0,0 +1,55 @@
...
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import replace_name
+
+
+PY2_TYPE = 'Text'
+PY3_TYPE = 'str'
```
asottile, line 16: I would inline these, they don't really give much as constants

tests/features/typing_text_test.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from pyupgrade._data import Settings
+from pyupgrade._main import _fix_plugins
+
+
+@pytest.mark.parametrize(
```
asottile, line 7: usually there's a set of noop tests here -- you probably want at least (1) no rewrite for < python 3 (2) unrelated names are not rewritten

asottile (commented):
otherwise looks good

asottile (approved):

MarcoGorelli:
Thank you so much for your review and approval!

I'm amazed by how well-structured your code is, I learned a ton just reading through it

Issue #397, opened 2021-02-10
Remove from __future__ import annotations on --py310plus


ShadowJonathan:
[3.10's changelog](https://docs.python.org/3.10/whatsnew/3.10.html#pep-563-postponed-evaluation-of-annotations-becomes-default) details how postponed evaluation of annotations becomes default, making annotations strings, instead of computed values.

`from __future__ import annotations` is imported in variable points, but since 3.10, it'll be possible to remove this import.

asottile:
3.10 has not been finalized yet so this won't be included until then

PR #402, opened 2021-03-03
Remove future annotations import with --py310-plus

cdce8p:
This PR adds `from __future__ import annotations` to the list of imports that should be removed (if `--py310-plus` is specified).

https://docs.python.org/3.9/library/__future__.html
https://bugs.python.org/issue38605

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ Availability:
 - `--py3-plus` will also remove `absolute_import` / `division` /
   `print_function` / `unicode_literals`
 - `--py37-plus` will also remove `generator_stop`
+- `--py310-plus` will also remove `annotations`
 
 ```diff
 -from __future__ import with_statement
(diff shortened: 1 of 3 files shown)


asottile:
https://github.com/asottile/pyupgrade/issues/397#issuecomment-776640442

cdce8p:
> #397 (comment)

Didn't saw that.

3.10 might not be finalized but it's included in the alpha releases and I wouldn't expect it to be removed now.
Besides `reorder_python_imports` already supports it: https://github.com/asottile/reorder_python_imports#removing-obsolete-__future__-imports So why not include it in `pyupgrade` as well?

wwuck:
@asottile are there any use cases for `__future__.annotations` in 3.14+?

asottile:
? this is a closed pr about python 3.10

PR #406, opened 2021-03-11
quoted annotations + `from __future__ import annotations` 

### LINKED ISSUE #255: quoted annotations + `from __future__ import annotations`
should remove the quotes

need to be careful of:
- `Literal['foo']`
- `Annotated` ([PEP 593](https://www.python.org/dev/peps/pep-0593/))

MarcoGorelli:
closes #255  (at least, attempts to)

### DIFF
--- a/pyupgrade/_plugins/typing_pep563.py
+++ b/pyupgrade/_plugins/typing_pep563.py
@@ -1,0 +1,168 @@
+import ast
+import functools
+import sys
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
+
+from tokenize_rt import Offset
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import Token
+from pyupgrade._data import TokenFunc
+
+
+def _supported_version(state: State) -> bool:
+    return (
+        state.settings.min_version >= (3, 10) or
+        'annotations' in state.from_imports['__future__']
+    )
+
+
+def _dequote(i: int, tokens: List[Token], *, new: str) -> None:
+    tokens[i] = tokens[i]._replace(src=new)
+
+
+def _get_name(node: ast.expr) -> str:
+    if isinstance(node, ast.Name):
+        return node.id
+    elif isinstance(node, ast.Attribute):
+        return node.attr
+    else:
+        raise AssertionError(f'expected Name or Attribute: {ast.dump(node)}')
+
+
+def _get_keyword_value(
+        keywords: List[ast.keyword],
+        keyword: str,
+) -> Optional[ast.expr]:
+    for kw in keywords:
+        if kw.arg == keyword:
+            return kw.value
+    else:
+        return None
+
+
+def _process_call(node: ast.Call) -> Iterable[ast.AST]:
+    name = _get_name(node.func)
+    args = node.args
+    keywords = node.keywords
+    if name == 'TypedDict':
+        if keywords:
+            for keyword in keywords:
+                yield keyword.value
+        elif len(args) != 2:  # garbage
+            pass
+        elif isinstance(args[1], ast.Dict):
+            yield from args[1].values
+        else:
+            raise AssertionError(f'expected ast.Dict: {ast.dump(args[1])}')
+    elif name == 'NamedTuple':
+        if len(args) == 2:
+            fields: Optional[ast.expr] = args[1]
+        elif keywords:
+            fields = _get_keyword_value(keywords, 'fields')
+        else:  # garbage
+            fields = None
+
+        if isinstance(fields, ast.List):
+            for elt in fields.elts:
+                if isinstance(elt, ast.Tuple) and len(elt.elts) == 2:
+                    yield elt.elts[1]
+        elif fields is not None:
+            raise AssertionError(f'expected ast.List: {ast.dump(fields)}')
+    elif name in {
+        'Arg',
+        'DefaultArg',
+        'NamedArg',
+        'DefaultNamedArg',
+        'VarArg',
+        'KwArg',
+    }:
+        if args:
+            yield args[0]
+        else:
+            keyword_value = _get_keyword_value(keywords, 'type')
+            if keyword_value is not None:
+                yield keyword_value
+
+
+def _process_subscript(node: ast.Subscript) -> Iterable[ast.AST]:
+    name = _get_name(node.value)
+    if name == 'Annotated':
+        if sys.version_info >= (3, 9):  # pragma: no cover (py39+)
+            node_slice: ast.expr = node.slice
+        elif isinstance(node.slice, ast.Index):  # pragma: no cover (<py39)
+            node_slice = node.slice.value
+        else:  # pragma: no cover (<py39)
+            pass  # unexpected slice type
+
+        if isinstance(node_slice, ast.Tuple):
+            if node_slice.elts:
+                yield node_slice.elts[0]
+        else:
+            raise AssertionError(f'expected ast.Tuple: {ast.dump(node_slice)}')
+    elif name != 'Literal':
+        yield node.slice
+
+
+def _replace_string_literal(
+        annotation: ast.expr,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    nodes: List[ast.AST] = [annotation]
+    while nodes:
+        node = nodes.pop()
+        if isinstance(node, ast.Call):
+            nodes.extend(_process_call(node))
+        elif isinstance(node, ast.Subscript):
+            nodes.extend(_process_subscript(node))
+        elif isinstance(node, ast.Str):
+            func = functools.partial(_dequote, new=node.s)
+            yield ast_to_offset(node), func
+        else:
+            for name in node._fields:
+                value = getattr(node, name)
+                if isinstance(value, ast.AST):
+                    nodes.append(value)
+                elif isinstance(value, list):
+                    nodes.extend(value)
+
+
+def _process_args(
+        args: Sequence[Optional[ast.arg]],
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    for arg in args:
+        if arg is not None and arg.annotation is not None:
+            yield from _replace_string_literal(arg.annotation)
+
+
+@register(ast.FunctionDef)
+def visit_FunctionDef(
+        state: State,
+        node: ast.FunctionDef,
+        parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+
+    yield from _process_args([node.args.vararg, node.args.kwarg])
+    yield from _process_args(node.args.args)
+    yield from _process_args(node.args.kwonlyargs)
+    yield from _process_args(getattr(node.args, 'posonlyargs', []))
+    if node.returns is not None:
+        yield from _replace_string_literal(node.returns)
+
+
+@register(ast.AnnAssign)
+def visit_AnnAssign(
+        state: State,
+        node: ast.AnnAssign,
+        parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+    yield from _replace_string_literal(node.annotation)
--- a/tests/features/typing_pep563_test.py
+++ b/tests/features/typing_pep563_test.py
@@ -1,0 +1,365 @@
+import sys
+
+import pytest
+
+from pyupgrade._data import Settings
+from pyupgrade._main import _fix_plugins
+
+
+@pytest.mark.parametrize(
+    ('s', 'version'),
+    (
+        pytest.param(
+            'from typing import Literal\n'
+            'x: "str"\n',
+            (2, 7),
+            id='not python 3.10+',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'x: Literal["foo", "bar"]\n',
+            (3,),
+            id='Literal',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'x = TypeVar("x", "str")\n',
+            (3,),
+            id='TypeVar',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'x = cast(x, "str")\n',
+            (3,),
+            id='cast',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'X = List["MyClass"]\n',
+            (3,),
+            id='Alias',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'X: MyCallable("X")\n',
+            (3,),
+            id='Custom callable',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'def foo(x, *args, **kwargs): ...\n',
+            (3,),
+            id='Untyped',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'def foo(*, inplace): ...\n',
+            (3,),
+            id='Kwonly, untyped',
+        ),
+    ),
+)
+def test_fix_typing_pep563_noop(s, version):
+    assert _fix_plugins(s, settings=Settings(min_version=version)) == s
+
+
+@pytest.mark.parametrize(
+    ('s', 'expected'),
+    (
+        pytest.param(
+            'from __future__ import annotations\n'
+            'def foo(var: "MyClass") -> "MyClass":\n'
+            '   x: "MyClass"\n',
+
+            'from __future__ import annotations\n'
+            'def foo(var: MyClass) -> MyClass:\n'
+            '   x: MyClass\n',
+
+            id='Simple annotation',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'def foo(*, inplace: "bool"): ...\n',
+
+            'from __future__ import annotations\n'
+            'def foo(*, inplace: bool): ...\n',
+
+            id='Kwonly, typed',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'def foo(*args: "str", **kwargs: "int"): ...\n',
+
+            'from __future__ import annotations\n'
+            'def foo(*args: str, **kwargs: int): ...\n',
+
+            id='Vararg and kwarg typed',
+        ),
+        pytest.param(
+            'from __future__ import annotations\n'
+            'x: Tuple["MyClass"]\n',
+
+            'from __future__ import annotations\n'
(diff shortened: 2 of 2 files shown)


pyupgrade/_token_helpers.py
```diff
@@ -434,3 +434,7 @@ def replace_name(i: int, tokens: List[Token], *, name: str, new: str) -> None:
             return
         j += 1
     tokens[i:j + 1] = [new_token]
+
+
+def dequote(i: int, tokens: List[Token], *, new: str) -> None:
+    tokens[i] = tokens[i]._replace(src=new)
```
asottile, line 440: if this is only used in one module it should live there

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+
+def _supported_version(state: State) -> bool:
+    return (
+        state.settings.min_version >= (3, 10) or (
+            'annotations' in state.from_imports['__future__']
+        )
+    )
+
+
+def replace_string_literal(
```
asottile, line 26: this should be underscore prefixed

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+def visit_AnnAssign(
+    state: State,
+    node: ast.AnnAssign,
+    parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+
+    to_replace: MutableMapping[ast.expr, str] = {}
+    replace_string_literal(node.annotation, to_replace)
```
asottile, line 84: this should return a mapping, not produce one as a side-effect

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+def visit_FunctionDef(
+    state: State,
+    node: ast.FunctionDef,
+    parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+    to_replace: MutableMapping[ast.expr, str] = {}
+    for i in node.args.args:
+        replace_string_literal(i.annotation, to_replace)
```
asottile, line 59: the rest of the code tries to avoid recursion here as it can easily bump into the limit on deeply nested constructs (this was one of the motivations for the plugin-based system)

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+    elif isinstance(annotation, (ast.Tuple, ast.List)):
+        for i in annotation.elts:
+            replace_string_literal(i, to_replace)
+    elif isinstance(annotation, ast.Index):  # pragma: no cover
+        # Not for Python 3.9
+        replace_string_literal(annotation.value, to_replace)
+    elif isinstance(annotation, ast.Subscript):
+        if isinstance(annotation.value, ast.Name):
+            if annotation.value.id != 'Literal':
+                replace_string_literal(annotation.slice, to_replace)
```
asottile, line 46: need to handle all the other special cases mentioned in the issue

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+
+def replace_string_literal(
+    annotation: Optional[Union[ast.expr, ast.slice]],
+    to_replace: MutableMapping[ast.expr, str],
+) -> None:
+    if isinstance(annotation, ast.Constant):
+        if isinstance(annotation.value, str):  # pragma: no cover
+            # Not for Python 3.6, Python 3.7
```
asottile, line 32: for version-specific no-covers, place them inside the pragma itself: `# pragma: no cover (py38+)`

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+            to_replace[annotation] = annotation.value
+    elif isinstance(annotation, ast.Str):  # pragma: no cover
+        # Only for Python 3.6, Python 3.7
+        to_replace[annotation] = annotation.s
+    elif isinstance(annotation, (ast.Tuple, ast.List)):
+        for i in annotation.elts:
```
asottile, line 38: `i` is not the right variable name (usually reserved for numbers), `el` is probably better

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,90 @@
...
+        replace_string_literal(i.annotation, to_replace)
+    for i in getattr(node.args, 'posonlyargs', []):  # pragma: no cover
+        # Only for Python3.8+
+        replace_string_literal(i.annotation, to_replace)
+    replace_string_literal(node.returns, to_replace)
+    for key, val in to_replace.items():
+        func = functools.partial(
+            dequote,
+            new=val,
+        )
```
asottile, line 70: this probably fits on one line

tests/features/typing_pep563_test.py
```diff
@@ -0,0 +1,148 @@
...
+        ),
+    ),
+)
+def test_fix_typing_pep563(s, expected):
+    ret = _fix_plugins(s, settings=Settings(min_version=(3, 7)))
+    assert ret == expected
+
+
+@pytest.mark.skipif(
+    sys.version_info.major != 3 or sys.version_info.minor != 8,
```
asottile, line 115: flake8-2020 should really yell at you here -- this logic is buggy don't do that do; `sys.version_info.major < (3, 8)` (think about what happens in 3.9 or 4.0)

also prefer `xfail` to `skip` -- it shouldn't crash and you'll get the coverage for the test without needing a no-cover

asottile (commented):
this needs to handle the cases outlined in the issue:

```
x: Annotated["Foo", "could really be anything here"]
x: TypedDict("Dct", {"foo": "bar"})  # there's another syntax too iirc
x: NamedTuple("X", [("foo", "bar")])  # there's another syntax too iirc
# etc.
```

MarcoGorelli:
Thanks @asottile for your review,

I can address most comments, but the one I'm stuck on is

> the rest of the code tries to avoid recursion here as it can easily bump into the limit on deeply nested constructs (this was one of the motivations for the plugin-based system)

I think I can't just do something like override `visit_Name` or `visit_Subscript` because then I won't know whether the node is part of an assignment (which, barring exceptions like `Literal`, can be dequoted) or a type alias (which can't be dequoted). Any suggestions on how to address that or should I close?

MarcoGorelli:
Thanks @asottile for your review - I've given this another go, and now there's a non-recursive solution, though it requires adding an extra attribute to `State`

pyupgrade/_data.py
```diff
@@ -103,6 +104,17 @@ def visit(
...
+                if isinstance(node.value, ast.Name):
+                    if node.value.id in {
+                        'Literal',
+                        'TypedDict',
+                        'NamedTuple',
+                        'Annotated',
+                    }:
+                        next_state = (
+                            next_state._replace(exclude_from_pep563=True)
+                        )
```
asottile, line 117: ah, you don't need to do this, I just meant implement your helper function non-recursively (use a local `list` / `deque` as a stack / queue)

MarcoGorelli: Thanks, I'll look into how to do that 

Though isn't this approach simpler, as it means only overriding Constant/Str rather than having to override FunctionDef and AnnAssign and then doing complicatedgymnastics to go through their attributes? The previous solution felt overly complex

asottile: continually rebuilding the state object is very slow -- and adding specific state for a single plugin is not good

MarcoGorelli:
Thanks @asottile for explaining

Regarding

> I just meant implement your helper function non-recursively (use a local list / deque as a stack / queue)

I've tried to address that in the last commit, is this what you meant?

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+    state: State,
+    node: ast.AnnAssign,
+    parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+    replacements = _replace_string_literal(node.annotation)
+    for replacement in replacements:
+        func = functools.partial(_dequote, new=replacement.new)
+        yield ast_to_offset(replacement.node), func
```
asottile, line 107: I think you can eliminate `Replacement` and get this to just `yield from _replace_string_literals(node.annotation)`

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+        return
+    replacements: MutableSequence[Replacement] = deque()
+    for args_name in node.args._fields:
+        args = getattr(node.args, args_name)
+        if args is None:
+            continue
+        for arg in args:
+            if hasattr(arg, 'annotation'):
+                value = getattr(arg, 'annotation')
+                replacements.extend(_replace_string_literal(value))
```
asottile, line 89: this is very weird, the args always have `.annotation` so you shouldn't need to `getattr` here

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+                    'NamedTuple',
+                    'TypedDict',
+                    'Annotated',
+                }:
+                    nodes.append(node.slice)
+        elif isinstance(node, ast.Constant):  # pragma: no cover (py37)
+            if isinstance(node.value, str):
+                replacements.append(Replacement(node, node.value))
+        elif isinstance(node, ast.Str):  # pragma: no cover (py38+)
+            replacements.append(Replacement(node, node.s))
```
asottile, line 61: `isinstance(node, ast.Str)` works in both versions -- I'd just do that for now

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+        node = nodes.pop()
+        if isinstance(node, ast.Subscript):
+            if isinstance(node.value, ast.Name):
+                if node.value.id not in {
+                    'Literal',
+                    'NamedTuple',
+                    'TypedDict',
+                    'Annotated',
+                }:
+                    nodes.append(node.slice)
```
asottile, line 56: this breaks for `typing.Annotated["foo", "bar"]`

also there are parts of those which are valid types that should be unquoted -- just not the whole thing

also for `NamedTuple` and `TypedDict` those aren't `Subscripts` those'll be calls

here's probably a useful testcase

```python
class D(TypedDict):
    E: TypedDict('E', {"foo": "int"})
```

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+    tokens[i] = tokens[i]._replace(src=new)
+
+
+def _replace_string_literal(
+    annotation: Optional[ast.expr],
+) -> Sequence[Replacement]:
+    nodes: MutableSequence[ast.AST] = deque()
+    if annotation is not None:
+        nodes.append(annotation)
+    replacements: MutableSequence[Replacement] = deque()
```
asottile, line 44: no reason for this to be a `deque` -- it is only `.append(...)`ed to

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _dequote(i: int, tokens: List[Token], *, new: str) -> None:
+    tokens[i] = tokens[i]._replace(src=new)
+
+
+def _replace_string_literal(
+    annotation: Optional[ast.expr],
+) -> Sequence[Replacement]:
+    nodes: MutableSequence[ast.AST] = deque()
```
asottile, line 41: no reason for this be a deque either, it just has `.pop()` and `.append()`

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+                    'Annotated',
+                }:
+                    nodes.append(node.slice)
+        elif isinstance(node, ast.Constant):  # pragma: no cover (py37)
+            if isinstance(node.value, str):
+                replacements.append(Replacement(node, node.value))
+        elif isinstance(node, ast.Str):  # pragma: no cover (py38+)
+            replacements.append(Replacement(node, node.s))
+        else:
+            for name in reversed(node._fields):
```
asottile, line 63: there's no need for `reversed` here -- it doesn't matter if the traversal is in-order or not (the callbacks end up being called in the proper order eventually)

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,107 @@
...
+
+    while nodes:
+        node = nodes.pop()
+        if isinstance(node, ast.Subscript):
+            if isinstance(node.value, ast.Name):
+                if node.value.id not in {
+                    'Literal',
+                    'NamedTuple',
+                    'TypedDict',
+                    'Annotated',
```
asottile, line 54: we should probably handle these as well from `mypy-extensions` (even though they haven't made it into standard typing yet) -- don't want to break people's code without an escape hatch: https://github.com/python/mypy_extensions/blob/dffd016aeb20851c6059b0077be033611f174500/mypy_extensions.py#L109-L136

tests/features/typing_pep563_test.py
```diff
@@ -0,0 +1,175 @@
...
+
+@pytest.mark.parametrize(
+    ('s', 'version'),
+    (
+        pytest.param(
+            'from typing import Literal\n'
+            'x: "str"\n',
+            (2, 7),
+            id='not python 3+',
+        ),
```
asottile, line 17: I think your test name here should be "not python3.10+" instead -- that's what you're actually testing in this case

tests/features/typing_pep563_test.py
```diff
@@ -0,0 +1,175 @@
...
+            '    var1: Tuple[MyClass],\n'
+            '    *,\n'
+            '    var2: Tuple[MyClass],\n'
+            ') -> Tuple[MyClass]:\n'
+            '   x: Tuple[MyClass]\n',
+
+            id='Tuple',
+        ),
+    ),
+)
```
asottile, line 172: probably not much of a need for `parametrize` here -- there's only one case.  also I'd stick to just testing what you intend to test here (cut down the argument list, also the `id` here is bogus so I'd just leave it out)

MarcoGorelli:
Thanks you @asottile for your excellent review, I've attempted to address all comments in the latest commit

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _process_call(node: ast.Call) -> Iterable[ast.AST]:
+    name = _get_name(node.func)
+    if name == 'TypedDict':
+        dict_ = node.args[1]
+        if isinstance(dict_, ast.Dict):
+            yield from dict_.values
+        else:
+            raise AssertionError(f'expected ast.Dict: {ast.dump(dict_)}')
```
asottile, line 46: TypedDict has two syntaxes, one with named arguments and one with a dictionary:

`TypedDict('foo', {'k': 'int'})` and `TypedDict('foo', k='int')`

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+        'DefaultNamedArg',
+        'VarArg',
+        'KwArg',
+    }:
+        if node.args:
+            yield node.args[0]
+        else:
+            for arg in node.keywords:
+                if arg.arg == 'type':
+                    yield arg.value
```
asottile, line 60: I think this also needs to handle the syntax for `NamedTuple` -- though I'm not sure those can be used anonymously so maybe not

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+            # Annotated can't be empty
+            pass
+    elif name not in {'Literal', 'NamedTuple'}:
+        yield node.slice
+
+
+def _replace_string_literal(
+    annotation: Optional[ast.expr],
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    nodes: List[ast.AST] = []
```
asottile, line 84: I think this is clearer (or even better would maybe be to not take `Optional` at all?)

```python
if annotation is None:
    return
nodes = [annotation]
```

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+            node_slice = node.slice.value
+        else:  # pragma: no cover (<py39)
+            pass  # unexpected slice type
+        if isinstance(node_slice, ast.Tuple):
+            yield node_slice.elts[0]
+        else:  # pragma: no cover
+            # Annotated can't be empty
+            pass
+    elif name not in {'Literal', 'NamedTuple'}:
+        yield node.slice
```
asottile, line 78: there are still types that need processing in `NamedTuple` -- also you can't have a `NamedTuple` slice -- that one is a `Call`

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+    if annotation is not None:
+        nodes.append(annotation)
+
+    while nodes:
+        node = nodes.pop()
+        if isinstance(node, ast.Call):
+            nodes.extend(_process_call(node))
+        elif isinstance(node, ast.Subscript):
+            nodes.extend(_process_subscript(node))
+        elif isinstance(node, ast.Str):  # pragma: no cover (py38+)
```
asottile, line 94: this shouldn't be no-covered

pyupgrade/_plugins/typing_pep563.py
```diff
@@ -0,0 +1,133 @@
...
+@register(ast.FunctionDef)
+def visit_FunctionDef(
+    state: State,
+    node: ast.FunctionDef,
+    parent: ast.AST,
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if not _supported_version(state):
+        return
+    for args_name in {'vararg', 'kwarg'}:
+        args = getattr(node.args, args_name)
```
asottile, line 115: I would prefer these to not use getattr -- you lose all the type safety there

MarcoGorelli:
Hey @asottile 

I'm still working on this, I just have a question - what's the expected output if someone uses incorrect syntax? e.g.
```python
x: NamedTuple('X', quox=[('foo', int)])
```
?
Should it be ignored, or should some error be thrown?

asottile:
garbage should be ignored

asottile (approved):
I fixed a few small things, hope you don't mind -- thanks for the patch!

MarcoGorelli:
> hope you don't mind

Not at all, I am _so_ grateful to you for the mentorship you provided me with, thank you so much 🙏!

PR #412, opened 2021-03-17
Rewrite open_mode even if mode is passed by name

MarcoGorelli:
On `master`,
```python
with open('t.py', mode='r') as fd:
    txt = fd.read()
```
won't get re-written

### DIFF
--- a/pyupgrade/_plugins/open_mode.py
+++ b/pyupgrade/_plugins/open_mode.py
@@ -1,6 +1,8 @@
 import ast
+import functools
 from typing import Iterable
 from typing import List
+from typing import NamedTuple
 from typing import Tuple
 
 from tokenize_rt import Offset
@@ -21,19 +23,28 @@ U_MODE_REMOVE_U = frozenset(('rUb', 'Urb', 'rbU', 'Ubr', 'bUr', 'brU'))
 U_MODE_REPLACE = U_MODE_REPLACE_R | U_MODE_REMOVE_U
 
 
-def _fix_open_mode(i: int, tokens: List[Token]) -> None:
+class FunctionArg(NamedTuple):
+    arg_idx: int
+    value: ast.expr
+
+
+def _fix_open_mode(i: int, tokens: List[Token], *, arg_idx: int) -> None:
     j = find_open_paren(tokens, i)
     func_args, end = parse_call_args(tokens, j)
-    mode = tokens_to_src(tokens[slice(*func_args[1])])
-    mode_stripped = mode.strip().strip('"\'')
+    mode = tokens_to_src(tokens[slice(*func_args[arg_idx])])
+    mode_stripped = mode.split('=')[-1]
+    mode_stripped = mode_stripped.strip().strip('"\'')
     if mode_stripped in U_MODE_REMOVE:
-        del tokens[func_args[0][1]:func_args[1][1]]
+        if arg_idx == 0:
+            del tokens[func_args[arg_idx][0]: func_args[arg_idx + 1][0]]
+        else:
+            del tokens[func_args[arg_idx - 1][1]:func_args[arg_idx][1]]
     elif mode_stripped in U_MODE_REPLACE_R:
         new_mode = mode.replace('U', 'r')
-        tokens[slice(*func_args[1])] = [Token('SRC', new_mode)]
+        tokens[slice(*func_args[arg_idx])] = [Token('SRC', new_mode)]
     elif mode_stripped in U_MODE_REMOVE_U:
         new_mode = mode.replace('U', '')
-        tokens[slice(*func_args[1])] = [Token('SRC', new_mode)]
+        tokens[slice(*func_args[arg_idx])] = [Token('SRC', new_mode)]
     else:
         raise AssertionError(f'unreachable: {mode!r}')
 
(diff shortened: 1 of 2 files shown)


pyupgrade/_plugins/open_mode.py
```diff
@@ -48,11 +49,21 @@ def visit_Call(
...
-    ):
-        yield ast_to_offset(node), _fix_open_mode
+            ):
+                yield ast_to_offset(node), _fix_open_mode
+        elif node.keywords:
+            mode = [i.value for i in node.keywords if i.arg == 'mode']
+            if (
+                mode and
+                isinstance(mode[0], ast.Str) and
+                mode[0].s in U_MODE_REMOVE.union(U_MODE_REPLACE)
```
asottile, line 65: this builds a set just to check containment (turns 2 O(1) operations into a slow O(N) operation)

pyupgrade/_plugins/open_mode.py
```diff
@@ -48,11 +49,21 @@ def visit_Call(
...
+            if (
                 node.args[1].s in U_MODE_REPLACE or
                 (len(node.args) == 2 and node.args[1].s in U_MODE_REMOVE)
-            )
-    ):
-        yield ast_to_offset(node), _fix_open_mode
+            ):
+                yield ast_to_offset(node), _fix_open_mode
+        elif node.keywords:
+            mode = [i.value for i in node.keywords if i.arg == 'mode']
```
asottile, line 61: would prefer `next((genexp), None)` instead of having to deal with a maybe-empty list you deal with a maybe-None value

asottile (commented):
this doesn't fix `open(foo, encoding='UTF-8', mode='r')`

pyupgrade/_plugins/open_mode.py
```diff
@@ -48,11 +49,24 @@ def visit_Call(
...
+                yield ast_to_offset(node), _fix_open_mode
+        elif node.keywords:
+            mode = next(
+                (i.value for i in node.keywords if i.arg == 'mode'),
+                None,
             )
-    ):
-        yield ast_to_offset(node), _fix_open_mode
+            if (
+                mode is not None and
```
asottile, line 66: now you don't need this check

pyupgrade/_plugins/open_mode.py
```diff
@@ -21,19 +23,35 @@
...
+    value: ast.expr
+
+
+def _fix_open_mode(i: int, tokens: List[Token], arg_idx: int) -> None:
     j = find_open_paren(tokens, i)
     func_args, end = parse_call_args(tokens, j)
-    mode = tokens_to_src(tokens[slice(*func_args[1])])
-    mode_stripped = mode.strip().strip('"\'')
+    if len(func_args) < 2:
+        return
```
asottile, line 35: when is this possible -- this should get filtered out from the ast step so we don't trigger this

pyupgrade/_plugins/open_mode.py
```diff
@@ -21,19 +23,35 @@
 U_MODE_REPLACE = U_MODE_REPLACE_R | U_MODE_REMOVE_U
 
 
-def _fix_open_mode(i: int, tokens: List[Token]) -> None:
+class FunctionArg(NamedTuple):
+    arg_idx: int
+    value: ast.expr
+
+
+def _fix_open_mode(i: int, tokens: List[Token], arg_idx: int) -> None:
```
asottile, line 31: let's make `arg_idx` named-only -- this is what I've been doing for other replace functions (makes it less error-prone)

pyupgrade/_plugins/open_mode.py
```diff
@@ -21,19 +23,35 @@
...
+    mode_stripped = mode.split('=')[-1]
+    mode_stripped = mode_stripped.strip().strip('"\'')
     if mode_stripped in U_MODE_REMOVE:
-        del tokens[func_args[0][1]:func_args[1][1]]
+        if arg_idx == 0:
+            next_token_slice = slice(*func_args[arg_idx + 1])
+            next_argument = tokens_to_src(tokens[next_token_slice])
+            tokens[next_token_slice] = [
+                Token('SRC', next_argument.lstrip(' ')),
+            ]
```
asottile, line 45: does this do weird stuff for multiline -- my guess is it doesn't come up much in practice -- it's also possible the existing code is broken for that too -- in which case just ignore me

pyupgrade/_plugins/open_mode.py
```diff
@@ -48,11 +66,39 @@ def visit_Call(
...
+                    mode.value.s in U_MODE_REPLACE
+                )
+            ):
+                func = functools.partial(
+                    _fix_open_mode,
+                    arg_idx=len(node.args) + mode.arg_idx,
+                )
+                yield ast_to_offset(node), func
+        else:
+            return
```
asottile, line 104: I think this can be left out

asottile (commented):
getting there!

pyupgrade/_plugins/open_mode.py
MarcoGorelli: I tried this out an it seemed to work fine with multilines - however, it's true that this doesn't come up much in practice (I've never seen it, I was just trying to cover all cases), so I've removed the extra logic as maybe it makes to leave whitespace formatting to other tools?

asottile (approved):

PR #416, opened 2021-04-02
Add opt-out for conversion of str.encode('utf-8') to str.encode()

hexagonrecursion:
pip does not want this https://github.com/pypa/pip/pull/9766#discussion_r606243093

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,9 @@ ur'\U0001f643'  # u'\U0001f643'
 
 ### `.encode()` to bytes literals
 
+Availability:
+- Unless `--keep-encoding` is passed.
+
 ```python
 'foo'.encode()           # b'foo'
 'foo'.encode('ascii')    # b'foo'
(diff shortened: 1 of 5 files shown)


asottile:
no thanks -- the existing options are for breakages only.  this is an opinionated tool without preferences

hexagonrecursion:
But pip prefers `str.encode('utf-8')` over `str.encode()` on the basis that explicit is better then implicit. What should I do?

Would you be interested in a PR that makes upupgrade leave `str.encode('utf-8')` alone and convert  `str.encode()` to `str.encode('utf-8')` unconditionally?

asottile:
> But pip prefers `str.encode('utf-8')` over `str.encode()` on the basis that explicit is better then implicit. What should I do?

the whole point of opinionated tools is to not have a discussion about things. if they don't like the opinions then don't use the tool.


> Would you be interested in a PR that makes upupgrade leave `str.encode('utf-8')` alone and convert `str.encode()` to `str.encode('utf-8')` unconditionally?

no, that's worse in every way

hexagonrecursion:
understood

PR #423, opened 2021-04-17
Drop universal wheel declaration

adamchainz:
This project supports Python 3 only now so the wheel should not be marked as universally supporting Pythons 2 and 3.


MarcoGorelli:
Hi @adamchainz - have you seen https://github.com/pre-commit/pre-commit/pull/1832 ?

adamchainz:
I did not and I apologise for this PR. I will go fix my own packages.

PR #430, opened 2021-05-02
CLN only replace argument token

MarcoGorelli:
Hi Anthony,

During stream yesterday you said that it might be better to do this by just replacing that one token - is this what you meant?


asottile (approved):
yep!

PR #433, opened 2021-05-06
replace stdout=PIPE and stderr=PIPE with capture_output=True

### LINKED ISSUE #398: Suggestion: Upgrade subprocess.run(stdout=subprocess.PIPE, stderr=subprocess.PIPE) to subprocess.run(capture_output=True) in --py37-plus

MarcoGorelli:
closes #398


tests/features/capture_output_test.py
```diff
@@ -0,0 +1,88 @@
...
+            'import subprocess\n'
+            'subprocess.run(["foo"], capture_output=True)\n',
+            id='stderr used before stdout',
+        ),
+        pytest.param(
+            'import subprocess\n'
+            'subprocess.run(stderr=subprocess.PIPE, args=["foo"], '
+            'stdout=subprocess.PIPE)\n',
+            'import subprocess\n'
+            'subprocess.run( args=["foo"], capture_output=True)\n',
```
asottile, line 81: this extra space seems wrong

pyupgrade/_plugins/capture_output.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def _replace_stdout(
+    tokens: List[Token],
+    func_args: Sequence[Tuple[int, int]],
+    stdout_arg_idx: int,
+) -> None:
+    tokens[slice(*func_args[stdout_arg_idx])] = [
+        Token('SRC', ' capture_output=True'),
+    ]
```
asottile, line 42: I don't think there's any benefit to having these be free functions

that said, I think there might be some benefit to having a "remove argument" and "replace argument" helper for the function rewriting (actually it might already exist but you'd have to poke around?)

pyupgrade/_plugins/capture_output.py
```diff
@@ -0,0 +1,97 @@
+import ast
```
asottile, line 1: maybe combine this file with the other subprocess args plugin(s)?

tests/features/capture_output_test.py
MarcoGorelli: This is the same as currently happens for `open(mode='r', file='foo')`, which becomes `open( file='foo')`

I'll add extra logic here and in `open` to remove this leading space if you like, although when I'd done that for `open` you'd commented that it seemed complicated and that this wouldn't come up much in practice (see this review comment: https://github.com/asottile/pyupgrade/pull/412#discussion_r597484470 ). IMO this is the same situtation - it's extremely rare that someone would call `subprocess.run(stderr=subprocess.PIPE, args=["foo"], stdout=subprocess.PIPE)`, and if someone's using pyupgrade they're likely also using a formatter which would take care of extra whitespaces

pyupgrade/_plugins/capture_output.py
MarcoGorelli: My thinking for having them is that I need to first operate on the second one, and then on the first one - otherwise the token indices will get shifted. I'll see if I can find another way though

tests/features/capture_output_test.py
MarcoGorelli: I've tried addressing this in the latest commit anyway

asottile (approved):

PR #436, opened 2021-05-11
Rewrite merge dictionaries

### LINKED ISSUE #321: Rewrite merge dictionaries
In Python 3.9 or greater (see  PEP-584):

```python
x = {"a":1}
y = {"b":2}
z = x | y
```

In Python 3.5 or greater:

```python
x = {"a":1}
y = {"b":2}
z = {**x, **y}
```

In Python 3.4 or greater:

```python
x = {"a":1}
y = {"b":2}
z = dict(**x, **y)
```

MarcoGorelli:
closes #321

### DIFF
--- a/tests/features/typing_584_test.py
+++ b/tests/features/typing_584_test.py
@@ -1,0 +1,97 @@
+import pytest
+
+from pyupgrade._data import Settings
+from pyupgrade._main import _fix_plugins
+
+
+@pytest.mark.parametrize(
+    ('s', 'version'),
+    (
+        pytest.param(
+            '{**a, **b}\n',
+            (3, 8),
+            id='<3.9',
+        ),
+        pytest.param(
+            '{"a": 0}\n',
+            (3, 9),
+            id='Dict without merge',
+        ),
+        pytest.param(
+            'x = {**a}\n',
+            (3, 9),
+            id='Merge of only one dict',
+        ),
+    ),
+)
+def test_fix_pep584_noop(s, version):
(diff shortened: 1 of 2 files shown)


tests/features/typing_584_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+            id='Simple dict rewrite',
+        ),
+        pytest.param(
+            'x = {**a}\n',
+
+            'x = a\n',
+
+            id='Merge of only one dict',
+        ),
```
asottile, line 42: this one isn't correct -- the _before_ is a shallow copy and the after is a direct reference

MarcoGorelli: Thanks, have updated!

Is there any drawback to using `Token('DUMMY', '')` or is `del tokens[i]` preferable?

asottile: ah so, the reason for those is if it were to process things in reverse, I think you can just `del` those in this case.

an aside: I just updated my repositories to replace "dummy" with "placeholder" (or other equivalent language) to make the code more inclusive

asottile (approved):

tests/features/typing_584_test.py
MarcoGorelli: I see - I was having issues here when using `del`. Not sure I understand why `del` works at all actually if it changes the number of tokens, but I'll make sense of it

>  I just updated my repositories to replace "dummy" with "placeholder" (or other equivalent language) to make the code more inclusive

Thanks for doing that 👍 

And thanks for your review/merge!

PR #440, opened 2021-05-14
Accept directories as arguments and recurse through them

abompard:
This commit makes it easier to use `pyupgrade` as a command line tool and not in the pre-commit hook.

Please feel free to adjust the language in the `README`, as English is not my first language.


asottile:
please search the issue tracker #150 #46 #339

Issue #257, opened 2020-02-10
upgrade list comprehension splat to generator expression splat


graingert:
`foo(*[fn(x) for x in items])` -> `foo(*(fn(x) for x in items))` for py >= 2.4

graingert:
and the same with unpacking

```
foo, bar, baz = [fn(x) for x in items]
```

asottile:
due to listcomp leak I think this can only safely be done for `--py3-plus`

zyv:
Hmmm, is there any reason why #961 was locked as off-topic o_O ?

The ruff people removed the rule because evidence was presented in astral-sh/ruff#12754 that it actually made code much slower. I was wondering why pyupgrade wanted to keep it, but the conversation was disabled.

Is it just a matter of style, performance notwithstanding, or is there more to it?

PR #444, opened 2021-05-16
Use generator to unpack list comprehension

MarcoGorelli:
Addresses part of #257

### DIFF
--- a/pyupgrade/_plugins/unpack_list_comprehension.py
+++ b/pyupgrade/_plugins/unpack_list_comprehension.py
@@ -1,0 +1,41 @@
+import ast
+import sys
+from typing import Iterable
+from typing import List
+from typing import Tuple
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_closing_bracket
+
+
+def _replace_list_comprehension(i: int, tokens: List[Token]) -> None:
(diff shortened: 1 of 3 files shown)


pyupgrade/_plugins/unpack_list_comprehension.py
```diff
@@ -0,0 +1,41 @@
...
+from pyupgrade._token_helpers import find_closing_bracket
+
+
+def _replace_list_comprehension(i: int, tokens: List[Token]) -> None:
+    if sys.version_info < (3, 8):  # pragma: no cover (py38+)
+        start = i - 1
+        while not (tokens[start].name == 'OP' and tokens[start].src == '['):
+            start -= 1
+    else:  # pragma: no cover (<py38)
+        start = i
```
asottile, line 23: I think there might be a helper for this somewhere in pyupgrade already? -- maybe try the set / dict comp rewrites?

MarcoGorelli: I've had a look at `_arg_token_index` and `victims` from `pyupgrade._token_helpers`, but it doesn't seem so easy to adapt them here.

They cover the case when the node yielded contains a function (such as `set`) which contains a comprehension. Here, on the other hand, the yielded node is the comprehension itself, and in <py38 I need to go _backwards_ to find the starting bracket (rather than forward to find the first element).

I'll take a closer look / have a harder think anyway

MarcoGorelli: Having looked at this further, I can't figure out how to use `victims` here...any hints? Else I'll get back to it later in the week, bit stuck on this

asottile: it's probably fine as is

asottile (approved):

PR #448, opened 2021-05-18
Use generator expressions in some built-in functions

### LINKED ISSUE #447: Use generator expressions in some built-in functions
From [PEP289](https://www.python.org/dev/peps/pep-0289/), in Python2.4+, some builtin functions could use generator expressions instead of list comprehensions, e.g.:

```diff
- sum([i for i in range(100)])
+ sum(i for i in range(100))
```
Likewise for `max` and `min`. Would this be a welcome addition?

MarcoGorelli:
closes #447

### DIFF
--- a/pyupgrade/_plugins/generator_expressions_pep289.py
+++ b/pyupgrade/_plugins/generator_expressions_pep289.py
@@ -1,0 +1,70 @@
+import ast
+from typing import Iterable
+from typing import List
+from typing import Tuple
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_closing_bracket
+from pyupgrade._token_helpers import find_comprehension_opening_bracket
+
+
+ALLOWED_FUNCS = frozenset((
+    'all',
+    'any',
+    'bytearray',
+    'bytes',
+    'frozenset',
+    'list',
+    'max',
+    'min',
+    'sorted',
+    'sum',
+    'tuple',
+))
+
+
+def _delete_list_comp_brackets(i: int, tokens: List[Token]) -> None:
+    start = find_comprehension_opening_bracket(i, tokens)
+    end = find_closing_bracket(tokens, start)
+    tokens[end] = Token('PLACEHOLDER', '')
+    tokens[start] = Token('PLACEHOLDER', '')
+
+
+def _replace_list_comp_brackets(i: int, tokens: List[Token]) -> None:
+    start = find_comprehension_opening_bracket(i, tokens)
+    end = find_closing_bracket(tokens, start)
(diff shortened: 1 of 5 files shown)


pyupgrade/_plugins/generator_expressions_pep289.py
```diff
@@ -0,0 +1,66 @@
...
+    (
+        'all',
+        'any',
+        'bytearray',
+        'bytes',
+        'frozenset',
+        'list',
+        'max',
+        'min',
+        'print',
```
asottile, line 28: print of a list is different than print of a generator

pyupgrade/_plugins/generator_expressions_pep289.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+ALLOWED_FUNCS = frozenset(
+    (
+        'all',
+        'any',
+        'bytearray',
+        'bytes',
+        'frozenset',
+        'list',
```
asottile, line 25: set should be added too

pyupgrade/_plugins/generator_expressions_pep289.py
```diff
@@ -0,0 +1,66 @@
...
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_closing_bracket
+from pyupgrade._token_helpers import find_comprehension_opening_bracket
+
+
+ALLOWED_FUNCS = frozenset(
+    (
```
asottile, line 19: you don't have to do the ugly thing black does here, you can use `frozenset((` and then `))`

tests/features/generator_expressions_pep289_test.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.parametrize(
+    ('s',),
+    (
+        pytest.param(
+            'sum((i for i in range(3)))',
+            id='Parenthesised generator expression',
+        ),
+        pytest.param(
+            'sum([i for i in range(2)], x=2)',
```
asottile, line 15: this one would be fine to rewrite, but it'd have to keep the parens

MarcoGorelli:
Thanks @asottile for your review, I've updated accordingly!

A note on `set` - this one's already covered by the other plugin which changes `set([i for i in range(3)])` into `{i for i in range(3)}`, so I don't think `set` should be included here as well - or should it?

pyupgrade/_plugins/generator_expressions_pep289.py
MarcoGorelli: Indeed, thanks - the "test" I wrote to make this list only checked the return types 😳 🤦

asottile:
> 
> 
> Thanks @asottile for your review, I've updated accordingly!
> 
> A note on `set` - this one's already covered by the other plugin which changes `set([i for i in range(3)])` into `{i for i in range(3)}`, so I don't think `set` should be included here as well - or should it?

oh tru

asottile (approved):

PR #477, opened 2021-06-23
rewrite cElementTree to ElementTree

### LINKED ISSUE #311: Rewrite xml.etree.cElementTree in Python 3.3+
In Python 3.3+, use `xml.etree.ElementTree` instead of `xml.etree.cElementTree`:

> _Changed in version 3.3_: This module will use a fast implementation whenever available. The `xml.etree.cElementTree` module is deprecated.

* https://docs.python.org/3/library/xml.etree.elementtree.html

---

It was planned to be removed in Python 3.9 and is currently removed in 3.9b1, but it may be pushed to 3.10 as there wasn't a deprecation warning:

* https://github.com/python/cpython/pull/19921
* https://bugs.python.org/issue36543

---

Some real examples of use with suggested rewrite:

```diff
+import xml.etree.cElementTree as ET
-import xml.etree.ElementTree as ET
```
```diff
-from xml.etree.cElementTree import XML
+from xml.etree.ElementTree import XML
```
```diff
-from xml.etree.cElementTree import Element
+from xml.etree.ElementTree import Element
```

It's often used with a fallback, would be good to ditch those too in 3.3+.

```diff
-try:
-    import xml.etree.cElementTree as default_etree
-except ImportError:
-    import xml.etree.ElementTree as default_etree
+import xml.etree.ElementTree as default_etree
```
```diff
-try:
-    from xml.etree import cElementTree as ETree
-except ImportError:
-    from xml.etree import ElementTree as ETree
+from xml.etree import ElementTree as ETree
```

MarcoGorelli:
closes #311 

there's not currently a `--py33-plus` flag, should that be added or is `--py36-plus` OK?

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,19 @@ Availability:
 ```
 
 
+### Rewrite `xml.etree.cElementTree` to `xml.etree.ElementTree`
+
+Availability:
+- `--py3-plus` is passed on the commandline.
+
+```diff
+-import xml.etree.cElementTree as ET
++import xml.etree.ElementTree as ET
+-from xml.etree.cElementTree import XML
++from xml.etree.ElementTree import XML
+```
+
+
 ### `typing.NamedTuple` / `typing.TypedDict` py36+ syntax
 
 Availability:
--- a/pyupgrade/_plugins/c_element_tree.py
+++ b/pyupgrade/_plugins/c_element_tree.py
@@ -1,0 +1,50 @@
+import ast
+from typing import Iterable
+from typing import List
+from typing import Tuple
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_token
+
+
+def _replace_celementtree_with_elementtree(
+        i: int,
+        tokens: List[Token],
+) -> None:
+    j = find_token(tokens, i, 'cElementTree')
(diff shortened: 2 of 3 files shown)


tests/features/c_element_tree_test.py
```diff
@@ -0,0 +1,57 @@
...
+            id='import multiple objects',
+        ),
+        pytest.param(
+            'import xml.etree.cElementTree as ET',
+            'import xml.etree.ElementTree as ET',
+            id='import with alias',
+        ),
+    ),
+)
+def test_fix_mock(s, expected):
```
asottile, line 56: this function name is copy pasted

README.md
```diff
@@ -496,6 +496,18 @@ if it would make the expression longer or if the substitution parameters are
 anything but simple names or dotted names (as this can decrease readability).
 
 
+### Rewrite `xml.etree.cElementTree` to `xml.etree.ElementTree`
+
+Availability:
+- `--py36-plus` is passed on the commandline.
```
asottile, line 502: this can be `--py3-plus` -- we have other older py3 target versions that are included there as well (mock, yield from, etc.)

pyupgrade/_plugins/c_element_tree.py
```diff
@@ -0,0 +1,51 @@
...
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_token
+
+
+def _replace_celementtree_with_elementtree(
+        i: int,
+        tokens: List[Token],
+) -> None:
+    j = find_token(tokens, i, 'cElementTree')
+    src = 'ElementTree'
```
asottile, line 21: inline this

asottile (approved):

PR #482, opened 2021-07-08
Upgrade to f-string get broken for Py3.6 in case await keyword is in

### LINKED ISSUE #481: Upgrade to f-string get broken for Py3.6 in case await keyword is in
First of all, thanks for the nice tool!

Here is my bug report. In case I use `--py36-plus`, I get:
```diff
-        print("Error while publishing certification: {}".format(await response.text()))
+        print(f"Error while publishing certification: {await response.text()}")
```

But, [`async` and `await` are not keywords in Py3.6. They started to be keywords in Py3.7](https://stackoverflow.com/a/43120967).
So, this change breaks Py3.6 support compatibility, since upper line doesn’t work with Py3.6. I would suggest to not upgrade to `f-string` in case there an `await` inside when using `--py36-plus`.

MarcoGorelli:
closes #481

### DIFF
--- a/pyupgrade/_main.py
+++ b/pyupgrade/_main.py
@@ -520,13 +520,22 @@ def _format_params(call: ast.Call) -> Set[str]:
     return params
 
 
+def _contains_await(node: ast.AST) -> bool:
+    for node_ in ast.walk(node):
+        if isinstance(node_, ast.Await):
+            return True
+    else:
+        return False
+
+
 class FindPy36Plus(ast.NodeVisitor):
-    def __init__(self) -> None:
+    def __init__(self, *, min_version: Version) -> None:
         self.fstrings: Dict[Offset, ast.Call] = {}
         self.named_tuples: Dict[Offset, ast.Call] = {}
         self.dict_typed_dicts: Dict[Offset, ast.Call] = {}
         self.kw_typed_dicts: Dict[Offset, ast.Call] = {}
         self._from_imports: Dict[str, Set[str]] = collections.defaultdict(set)
+        self.min_version = min_version
 
     def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
         if node.level == 0 and node.module in {'typing', 'typing_extensions'}:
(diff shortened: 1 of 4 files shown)


pyupgrade/_main.py
```diff
@@ -591,7 +592,14 @@ def visit_Call(self, node: ast.Call) -> None:
                     if not candidate:
                         i += 1
             else:
-                self.fstrings[ast_to_offset(node)] = node
+                if (
+                    self.min_version == (3, 6) and
+                    not (
+                        len(node.args) == 1 and
```
MarcoGorelli, line 598: this isn't right

pyupgrade/_main.py
```diff
@@ -758,13 +763,13 @@ def _typed_class_replacement(
...
 
-def _fix_py36_plus(contents_text: str) -> str:
+def _fix_py36_plus(contents_text: str, *, min_version: Version) -> str:
     try:
         ast_obj = ast_parse(contents_text)
     except SyntaxError:
         return contents_text
 
-    visitor = FindPy36Plus()
+    visitor = FindPy36Plus(min_version=min_version)
```
MarcoGorelli, line 772: 🤔  I think this might count as a boolean trap, but rewriting `Py36Plus` as `Py37Plus` just for a small difference struck me as duplicative

(of course, no offence taken if you want to close this to get your own fix out quickly)

pyupgrade/_main.py
```diff
@@ -591,7 +592,11 @@ def visit_Call(self, node: ast.Call) -> None:
                     if not candidate:
                         i += 1
             else:
-                self.fstrings[ast_to_offset(node)] = node
+                if (
+                    self.min_version > (3, 6) or
+                    not any(isinstance(arg, ast.Await) for arg in node.args)
```
asottile, line 597: this is insufficient as it could be something like `1 + await f`

MarcoGorelli: gotcha - OK, will push a fix now

pyupgrade/_main.py
```diff
@@ -591,6 +592,19 @@ def visit_Call(self, node: ast.Call) -> None:
...
+                        node_ = nodes.pop()
+                        if isinstance(node_, ast.Await):
+                            self.generic_visit(node)
+                            return
+                        for name in node_._fields:
+                            value = getattr(node_, name)
+                            if isinstance(value, ast.AST):
+                                nodes.append(value)
+                            elif isinstance(value, list):
+                                nodes.extend(value)
```
MarcoGorelli, line 607: might need to only do this for elements of `value` which as instances of `ast.AST`, I'll check later

asottile (approved):

PR #484, opened 2021-07-10
Add defensive check for pep563

MarcoGorelli:
I haven't been able to find a test case which would hit the `else` block here, but if there was one, then `nodes` would contain an object which isn't `ast.AST` and so wouldn't have a `_fields` attribute. Is there a test which could be added, or is a defensive check like this OK?

### DIFF
--- a/pyupgrade/_plugins/typing_pep563.py
+++ b/pyupgrade/_plugins/typing_pep563.py
@@ -129,7 +129,13 @@ def _replace_string_literal(
                 if isinstance(value, ast.AST):
                     nodes.append(value)
                 elif isinstance(value, list):
-                    nodes.extend(value)
+                    for value_ in value:
+                        if isinstance(value_, ast.AST):
+                            nodes.append(value_)
+                        else:
+                            raise AssertionError(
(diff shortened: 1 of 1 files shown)


asottile:
is the thing that's being defended against possible? if not, no need to write special code for it

MarcoGorelli:
I just worried that it was possible, just not something I'd considered when I'd made the pep563 PR

In `_data.py` you do this:

https://github.com/asottile/pyupgrade/blob/513fe79284cd8bae3274563c42c2df226381f6fe/pyupgrade/_data.py#L110-L113

, but there you could have that `value` isn't an `ast.AST` if you have
```
global a, b
```
, where the `astpretty` would give
```
Module(
    body=[Global(lineno=1, col_offset=0, end_lineno=1, end_col_offset=11, names=['a', 'b'])],
    type_ignores=[],
)
```
(so, `.names` would be a list of non-`ast.AST`s)

Another example is
```
{**{}}
```

Are there such cases in the context of annotations?

asottile:
I think neither of those are possible in annotations (?)

MarcoGorelli:
Cool, let's leave this then

PR #491, opened 2021-07-15
 async generator breakage

### LINKED ISSUE #489: async generator breakage
pyupgrade 2.21.0, python 3.8.0

```python
import asyncio


async def agen():
    for x in range(5):
        yield x


async def main():
    x = tuple([i ** 2 async for i in agen()])
    print(x)


asyncio.run(main())
```
```shellsession
$ python t2.py
(0, 1, 4, 9, 16)
```

pyupgrade makes this change:
```diff
--- t2.py.orig	2021-07-15 22:19:32.201257932 +0300
+++ t2.py	2021-07-15 22:23:54.089120663 +0300
@@ -10 +10 @@
-    x = tuple([i ** 2 async for i in agen()])
+    x = tuple(i ** 2 async for i in agen())
```

...which breaks:
```shellsession
$ python t2.py
  File "t2.py", line 14, in <module>
    asyncio.run(main())
[...]
    x = tuple(i ** 2 async for i in agen())
TypeError: 'async_generator' object is not iterable
```

MarcoGorelli:
closes #489

### DIFF
--- a/tests/features/generator_expressions_pep289_test.py
+++ b/tests/features/generator_expressions_pep289_test.py
@@ -20,6 +20,14 @@ from pyupgrade._main import _fix_plugins
             '"".join([[i for _ in range(2)] for i in range(3)])\n',
             id='string join (left alone for perf reasons)',
         ),
+        pytest.param(
+            'async def foo():\n'
+            '    for i in range(3):\n'
+            '        yield i\n'
+            'async def bar():\n'
+            '    sum([i async for i in foo()])\n',
+            id='Contains async',
+        ),
     ),
 )
 def test_fix_generator_expressions_noop(s):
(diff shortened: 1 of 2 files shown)


tests/features/generator_expressions_pep289_test.py
```diff
@@ -20,6 +20,10 @@
             '"".join([[i for _ in range(2)] for i in range(3)])\n',
             id='string join (left alone for perf reasons)',
         ),
+        pytest.param(
+            'sum([i async for i in foo()])\n',
```
asottile, line 24: this test doesn't fail before the change in python3.6 (this is a syntax error: `async for outside of async def`)

asottile (approved):

PR #494, opened 2021-07-16
rewrite six.moves.range

### LINKED ISSUE #340: upgrade six.moves.range to range (and other builtins)
before:

```python
import six.moves
from six.moves import range


def foo():
    return six.moves.range(1), range(1)
```

after:

```python
import six.moves


def foo():
    return six.moves.range(1), range(1)
```

expected:
```python
def foo():
    return range(1), range(1)
```

MarcoGorelli:
closes #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,8 @@ six.exec_(c, g, l)        # exec(c, g, l)
 six.advance_iterator(it)  # next(it)
 six.next(it)              # next(it)
 six.callable(x)           # callable(x)
+six.moves.range(x)        # range(x)
+six.moves.xrange(x)       # range(x)
 
 from six import text_type
 text_type                 # str
--- a/pyupgrade/_data.py
+++ b/pyupgrade/_data.py
@@ -48,6 +48,7 @@ RECORD_FROM_IMPORTS = frozenset((
     'mmap',
     'select',
     'six',
+    'six.moves',
     'socket',
     'subprocess',
     'sys',
--- a/pyupgrade/_plugins/six_simple.py
+++ b/pyupgrade/_plugins/six_simple.py
@@ -68,6 +68,16 @@ def visit_Attribute(
 
         func = functools.partial(replace_name, name=node.attr, new=new)
         yield ast_to_offset(node), func
+    elif (
+            state.settings.min_version >= (3,) and
+            isinstance(node.value, ast.Attribute) and
+            isinstance(node.value.value, ast.Name) and
+            node.value.value.id == 'six' and
+            node.value.attr == 'moves' and
+            node.attr in {'xrange', 'range'}
+    ):
+        func = functools.partial(replace_name, name=node.attr, new='range')
+        yield ast_to_offset(node), func
 
 
 @register(ast.Name)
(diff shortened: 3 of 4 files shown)


asottile:
there's quite a few more that should get handled probably: https://github.com/asottile/reorder_python_imports/blob/006b2ee37e304bdcb7b3c7ea6a0b1079c9cf999e/reorder_python_imports.py#L651-L728

MarcoGorelli:
I'm not sure I see how - those in `REMOVALS` are already handled here

https://github.com/asottile/pyupgrade/blob/ca5d7df35d0fa42196598df0f687740354f0fdab/pyupgrade/_main.py#L414

and those in `REPLACES` would require extra imports. E.g. I can't rewrite
```python
six.moves.BaseHTTPServer
```
to
```
http.server
```
because that would require `http` to have been imported, which as far as I understand is out of scope for pyupgrade

asottile:
ah I guess I misunderstood what this PR was aiming at, the `REPLACES` is only for `from` imports -- can always do that in another PR

pyupgrade/_plugins/six_moves.py
```diff
@@ -0,0 +1,53 @@
...
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import replace_name
+
+
+NAMES = {
+    'range': 'range',
+    'xrange': 'range',
+}
```
asottile, line 18: since they're currently always `range` -- this can just be inlined as `'range'` right?

pyupgrade/_plugins/six_moves.py
```diff
@@ -0,0 +1,53 @@
+import ast
```
asottile, line 1: should this get added to the existing `six` module ?

MarcoGorelli:
> ah I guess I misunderstood what this PR was aiming at, the `REPLACES` is only for `from` imports -- can always do that in another PR

ah, so to rewrite e.g.
```
from six import StringIO
from six.moves.builtins import bool
```
to
```
from io import StringIO
from builtins import bool
```
?

Sure, sounds good, I'll try that in a separate PR (and'll update this one later)

asottile:
> 
> 
> > ah I guess I misunderstood what this PR was aiming at, the `REPLACES` is only for `from` imports -- can always do that in another PR
> 
> ah, so to rewrite e.g.
> 
> ```
> from six import StringIO
> from six.moves.builtins import bool
> ```
> 
> to
> 
> ```
> from io import StringIO
> from builtins import bool
> ```
> 
> ?
> 
> Sure, sounds good, I'll try that in a separate PR (and'll update this one later)

yep!  though actually the `six.moves.builtins` ones should just be dropped I imagine

asottile (approved):

graingert:
Hyped for Monday already

MarcoGorelli:
> > > ah I guess I misunderstood what this PR was aiming at, the `REPLACES` is only for `from` imports -- can always do that in another PR
> > 
> > 
> > ah, so to rewrite e.g.
> > ```
> > from six import StringIO
> > from six.moves.builtins import bool
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > from io import StringIO
> > from builtins import bool
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> > Sure, sounds good, I'll try that in a separate PR (and'll update this one later)
> 
> yep! though actually the `six.moves.builtins` ones should just be dropped I imagine

Just to clarify - would this just be a duplication of what `reorder-python-imports` does?

asottile:
for the most part, yeah

PR #393, opened 2021-02-05
lru_cache(maxsize=None) => cache

sbrugman:
Contributing a plugin for https://github.com/asottile/pyupgrade/issues/348

PR #495, opened 2021-07-20
Rewrite lru_cache with maxsize=None and typed=False

MarcoGorelli:
xref https://github.com/asottile/pyupgrade/pull/393#discussion_r572895574 (@mxr is this what you were suggesting?)

### DIFF
--- a/pyupgrade/_plugins/lru_cache.py
+++ b/pyupgrade/_plugins/lru_cache.py
@@ -2,6 +2,7 @@ import ast
 import functools
 from typing import Iterable
 from typing import List
+from typing import Optional
 from typing import Tuple
 
 from tokenize_rt import Offset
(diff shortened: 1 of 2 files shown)


pyupgrade/_plugins/lru_cache.py
```diff
@@ -47,13 +47,27 @@ def visit_Call(
             node.func.attr == 'lru_cache' and
             isinstance(node.func.value, ast.Name) and
             node.func.value.id == 'functools' and
-            not node.args and
-            len(node.keywords) == 1 and
-            node.keywords[0].arg == 'maxsize' and
-            isinstance(node.keywords[0].value, ast.NameConstant) and
-            node.keywords[0].value.value is None
+            not node.args
```
mxr, line 50: I think it would be cleaner to rewrite this as `not node.args and _eligible(node.keywords)` and the implementation of `_eligible` would check the kwargs to ensure they fit the criteria for rewriting, which is something like:

1) 1 kwarg and it's `maxsize=None`
2) 2 kwargs and they're `maxsize=None, typed=False` in any order

mxr: There's also a broader question of whether or not we want to do this rewrite (I think it makes sense but will defer to Anthony)

mxr:
Thanks for the PR - this is indeed what I was suggesting!

pyupgrade/_plugins/lru_cache.py
mxr:
```python

def _eligible(keywords: List[ast.keyword]) -> bool:
    if len(keywords) == 1:
        return _is_literal_kwarg(keywords[0], 'maxsize', None)
    elif len(keywords) == 2:
        return (
            (
                _is_literal_kwarg(keywords[0], 'maxsize', None) and
                _is_literal_kwarg(keywords[1], 'typed', False)
            ) or
            (
                _is_literal_kwarg(keywords[1], 'maxsize', None) and
                _is_literal_kwarg(keywords[0], 'typed', False)
            )
        )
    else:
        return False


def _is_literal_kwarg(
        keyword: ast.keyword, name: str, value: Optional[bool],
) -> bool:
    return (
        keyword.arg == name and
        isinstance(keyword.value, ast.NameConstant) and
        keyword.value.value is value
    )

```

MarcoGorelli: Indeed, that requires a couple of extra tests to get the coverage up, but I agree that it's cleaner / more readable. Thanks! I've added you as co-author to the commit

asottile (approved):
(will merge when the six one is ready!)

PR #499, opened 2021-07-25
replace type(''), type(b''), and type(0) with str, bytes, and int

### LINKED ISSUE #498: replace `type('')`, `type(b'')`, and `type(0)` with `str`, `bytes`, and `int`
it was relatively common when porting to use these in place of the switched-on types

MarcoGorelli:
closes #498

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,23 @@ Availability:
 ```
 
 
+### Rewrite `type` of primitive
+
+Availability:
+- `--py3-plus` is passed on the commandline.
+
+```diff
+-type('')
++str
+-type(b'')
++bytes
+-type(0)
++int
+-type(0.)
++float
+```
+
+
 ### `typing.NamedTuple` / `typing.TypedDict` py36+ syntax
 
 Availability:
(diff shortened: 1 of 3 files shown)


t.py
```diff
@@ -0,0 +1,2 @@
+type(True)
```
asottile, line 1: he he he

MarcoGorelli: caught me :flushed:

MarcoGorelli:
this'll do the rewrite even if the string isn't empty, e.g. `type("foo")` - is that OK or does it need excluding?

asottile:
> 
> 
> this'll do the rewrite even if the string isn't empty, e.g. `type("foo")` - is that OK or does it need excluding?

yep! that's fine

README.md
```diff
@@ -458,6 +458,23 @@ Availability:
...
+-type(0.)
++float
+```
+
+
+### Rewrite `type` of primitive
+
+Availability:
+- `--py3-plus` is passed on the commandline.
+
```
asottile, line 477: looks like the header is out of place

MarcoGorelli: indeed, thanks!

asottile (approved):

PR #509, opened 2021-08-03
support directories in main targets

eviljeff:
Patch to add support for recursively upgrading `.py` files in folders too, written on the back of mozilla/addons-server#17571


eviljeff:
Hi @asottile was this PR unintentionally closed or should I have done something else before opening it?  (I can look into the CI fails - looks like Linux is returning the files in a different order to Windows)

asottile:
intentionally closed, see them others which are closed with the same feature that I intentionally do not support

PR #520, opened 2021-08-17
Rewrite unittest deprecated aliases

### LINKED ISSUE #519: unittest deprecated aliases
`unittest` has a bunch of deprecated aliases: https://docs.python.org/3.9/library/unittest.html#deprecated-aliases

Any use triggers deprecation warnings in one's test suite:

```
/.../test_example.py:6: DeprecationWarning: Please use assertEqual instead.
  self.assertEquals(1, 1)
```

It would be nice to rewrite those.

adamchainz:
Fixes #519.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,25 @@ print(("foo"))                  # print("foo")
 
 [python-modernize/python-modernize#178]: https://github.com/python-modernize/python-modernize/issues/178
 
+### unittest deprecated aliases
+
+Rewrites [deprecated unittest method aliases](https://docs.python.org/3/library/unittest.html#deprecated-aliases) to their non-deprecated forms.
+
+Availability:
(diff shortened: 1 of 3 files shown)


adamchainz:
I was thinking of trying a little harder to detect whether we're inside a `unittest` TestCase, but it's hard since users often use subclasses of `TestCase` e.g. via Django, or project specific.

MarcoGorelli (changes requested):
Hey @adamchainz 

Quick comment before Anthony's review - I expect you'll be asked to provide `noop` test cases for when pyupgrade isn't expected to do a rewrite (see other test files for examples)

adamchainz:
Very true. One of my tests was even a 'noop' test already, I just didn't write it as such.

asottile (approved):

PR #530, opened 2021-09-10
ENH: expand versioned_branches feature to Python 3 minor version comparison (<, >, <=, >= with else)

### LINKED ISSUE #488: Feature idea/question: cleanup conditionals on sys.version_info
This is more of a question at this point: I wonder if it _would_ be possible to do this with pyupgrade.
If it is, I'll attempt to write a patch, otherwise you can just close this !

## input

assuming we run pyupgrade with the `--py39` flag
```python
import sys

if sys.version_info < (3, 9):
    print("this should be cleaned up")
else:
    print("this is the message we should keep (dedent)")

if sys.version_info > (3, 8):
    print("this is the message we should keep (dedent)")
else:
    print("this should be cleaned up")

if sys.version_info >= (3, 8):
    print("this is the message we should keep (dedent)")
else:
    print("this should be cleaned up")

if sys.version_info <= (3, 9):
    print("nothing to do here")
else:
    print("both branches are still valid")
```
## output
```python
import sys

print("this is the message we should keep (dedent)")

print("this is the message we should keep (dedent)")

print("this is the message we should keep (dedent)")

if sys.version_info <= (3, 9):
    print("nothing to do here")
else:
    print("both branches are still valid")
```

neutrinoceros:
Fix #488
It should cover Python 3.x branches (where x is >=0) using operators `<`, `<=`, `>`, `>=`, and assuming
that an `else` statement exists.

A couple remarks:
- this is my very first contribution to anything that uses `ast`, so I'm kind of astonished that my own tests are passing, but I'll gladly take any suggestions on how to make this better
- I ended up kinda "hacking" existing functions (namely `_fix_py3_block`, `_fix_py3_block_else` and `_fix_py2_block`) in the sense that I'm using them without modifications in a different fashion compared to their original intended usage. I suppose they ought to be renamed to better reflect what they actually do here (fix and if/else block and keep the top or lower half ?), but I leave this as a question to the maintainer
- I didn't cover the `==` operator. I can do it if you feel it's necessary, but I thought I'd ask for feedback first.

So... what do you think ? anything obvious that I'm not covering in the tests yet ?

### DIFF
--- a/pyupgrade/_plugins/versioned_branches.py
+++ b/pyupgrade/_plugins/versioned_branches.py
@@ -14,6 +14,7 @@ from pyupgrade._ast_helpers import is_name_attr
 from pyupgrade._data import register
 from pyupgrade._data import State
 from pyupgrade._data import TokenFunc
+from pyupgrade._data import Version
 from pyupgrade._token_helpers import Block
 
 
@@ -77,6 +78,7 @@ def _eq(test: ast.Compare, n: int) -> bool:
 def _compare_to_3(
     test: ast.Compare,
     op: Union[Type[ast.cmpop], Tuple[Type[ast.cmpop], ...]],
+    minor: int = 0,
 ) -> bool:
     if not (
             isinstance(test.ops[0], op) and
@@ -87,9 +89,11 @@ def _compare_to_3(
         return False
 
     # checked above but mypy needs help
-    elts = cast('List[ast.Num]', test.comparators[0].elts)
+    ast_elts = cast('List[ast.Num]', test.comparators[0].elts)
+    # padding a 0 for compatibility with (3,) used as a spec
+    elts = tuple(e.n for e in ast_elts) + (0,)
 
-    return elts[0].n == 3 and all(n.n == 0 for n in elts[1:])
+    return elts[:2] == (3, minor) and all(n == 0 for n in elts[2:])
 
 
 @register(ast.If)
@@ -98,8 +102,16 @@ def visit_If(
         node: ast.If,
         parent: ast.AST,
 ) -> Iterable[Tuple[Offset, TokenFunc]]:
+
+    min_version: Version
+    if state.settings.min_version == (3,):
+        min_version = (3, 0)
+    else:
+        min_version = state.settings.min_version
+    assert len(min_version) >= 2
+
     if (
-            state.settings.min_version >= (3,) and (
+            min_version >= (3,) and (
                 # if six.PY2:
                 is_name_attr(node.test, state.from_imports, 'six', ('PY2',)) or
                 # if not six.PY3:
@@ -114,6 +126,7 @@ def visit_If(
                     )
                 ) or
                 # sys.version_info == 2 or < (3,)
+                # or < (3, n) or <= (3, n) (with n<m)
                 (
                     isinstance(node.test, ast.Compare) and
                     is_name_attr(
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Also; should I add docs too ?

neutrinoceros:
Ah, I see that this actually only gets 3/4 cases right from the associated issue's example, so this is definitely not complete, but right now it's bedtime for me, I'll add and fix more tests tomorrow !

neutrinoceros:
Added more tests, and edited the OP in the attached issue (I think the 4th example was incorrect there _and_ in my results last night). Now every case -that I can think of- works.

neutrinoceros:
Alright, after a couple iterations (more tests and bugfixes), this is now in a state that covers every case I can currently think of correctly 🎉

pyupgrade/_plugins/versioned_branches.py
```diff
@@ -98,6 +106,11 @@ def visit_If(
         node: ast.If,
         parent: ast.AST,
 ) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if state.settings.min_version >= (3,) and (
+        len(state.settings.min_version) >= 2
+    ):
+        py3_minor = state.settings.min_version[1]
```
asottile, line 112: this makes a sometimes-present variable which is a bit of a hazard

the whole patch can be very much simplified by changing `(3,)` => `(3, 0)` and then you don't need any special logic

neutrinoceros: > this makes a sometimes-present variable which is a bit of a hazard

point taken.

> the whole patch can be very much simplified by changing (3,) => (3, 0) and then you don't need any special logic

Not so sure about that: `state.settings.min_version` isn't guaranteed to have a second item (with `--py3-plus` it's just `(3,)`), but `(3,) >= (3, 0)` evaluates to `False`. Should I change this ?

More generally I _think_ you're right that it can be simplified, at least it feels a lot like it, I just don't see how right now.

asottile: I think you can just set:
```python
if state.settings.min_version == (3,):
    min_version = (3, 0)
else:
    min_version = state.settings.min_version
```

or something of the sort

neutrinoceros:
@asottile I was able to simplify the logic in `_compare_to_3`, I think it doesn't get much simpler from here. I also added some docs. Now ready for your next (final ?) review :)

neutrinoceros:
added a "simplification" that avoids repeating a check when the minor version is > 0, but I'm not sure it's worth the obfuscation. Leaving it as an isolated commit so it can be easily pruned or squashed.

pyupgrade/_plugins/versioned_branches.py
```diff
@@ -124,15 +137,19 @@ def visit_If(
                     ) and
                     len(node.test.ops) == 1 and (
                         _eq(node.test, 2) or
-                        _compare_to_3(node.test, ast.Lt)
+                        _compare_to_3(node.test, ast.Lt, min_version[1]) or
+                        any(
+                            _compare_to_3(node.test, (ast.Lt, ast.LtE), minor)
+                            for minor in range(min_version[1])
+                        )
```
asottile, line 144: I think this `any` block could be eliminated and subsumed into the `_compare_to` function

right now it's linear on the version being compared which could potentially be bad (?)

I'm going to merge this as is and we can always fix it later?

asottile (approved):

pyupgrade/_plugins/versioned_branches.py
neutrinoceros: Sounds like a plan :)

PR #533, opened 2021-09-13
BUG: fix broken tests for versioned_branches feature

neutrinoceros:
Followup to #530, thanks to [a comment from @graingert](https://github.com/asottile/pyupgrade/commit/31546d2cdf4a7647df46b0e9c4b771efe123ea3b#r56396117)

Actually the bug now fails, revealing a bug in #530 ? I'll try to fix this promptly

### DIFF
--- a/tests/features/versioned_branches_test.py
+++ b/tests/features/versioned_branches_test.py
@@ -545,6 +545,15 @@ def test_fix_py3_only_code(s, expected):
             '3+6\n',
             id='from sys import version_info, <= (3, 5)',
         ),
+        pytest.param(
+            'import sys\n'
+            'if sys.version_info >= (3, 6):\n'
+            '    pass',
+
+            'import sys\n'
+            'pass',
+            id='sys.version_info >= (3, 6), noelse',
+        ),
     ),
 )
 def test_fix_py3x_only_code(s, expected):
@@ -555,15 +564,6 @@ def test_fix_py3x_only_code(s, expected):
 @pytest.mark.parametrize(
     's',
     (
-        # we timidly skip `if` without `else` as it could cause a SyntaxError
-        'import sys'
-        'if sys.version_info >= (3, 6):\n'
-        '    pass',
-        # here's the case where it causes a SyntaxError
-        'import sys'
-        'if True'
-        '    if sys.version_info >= (3, 6):\n'
-        '        pass\n',
         # both branches are still relevant in the following cases
         'import sys\n'
         'if sys.version_info > (3, 7):\n'


neutrinoceros:
@asottile !

pyupgrade/_plugins/versioned_branches.py
```diff
@@ -188,5 +188,10 @@ def visit_If(
...
         if node.orelse and not isinstance(node.orelse[0], ast.If):
             yield ast_to_offset(node), _fix_py3_block_else
-        elif not node.orelse:
+        elif not node.orelse and min_version == (3, 0):
+            # allow >= (3,) and > (3,) to be dedented even without
+            # an else clause, as a special cases
             yield ast_to_offset(node), _fix_py3_block
+        else:
+            # don't try to fix, e.g., >= (3, 6) if there's no else clause
+            yield ast_to_offset(node), lambda i, tokens: None
```
asottile, line 197: we shouldn't put noop callbacks into the callback machinery -- I think the original logic was fine here (?)

tests/features/versioned_branches_test.py
```diff
@@ -556,12 +556,12 @@ def test_fix_py3x_only_code(s, expected):
...
         # we timidly skip `if` without `else` as it could cause a SyntaxError
-        'import sys'
+        'import sys\n'
         'if sys.version_info >= (3, 6):\n'
         '    pass',
         # here's the case where it causes a SyntaxError
-        'import sys'
-        'if True'
+        'import sys\n'
+        'if True\n'
```
asottile, line 564: oh yeah this'll need a `:` too

neutrinoceros: good catch

pyupgrade/_plugins/versioned_branches.py
neutrinoceros: actually the result with the original logic for that one broken test is
```diff
import sys
if True:
-    if sys.version_info >= (3, 6):
-        pass
+   pass
```
so that sounds about right to me, I guess this test was misplaced and shouldn't have been in the "noop" zone at all ?
I'll push to revert this.

tests/features/versioned_branches_test.py
```diff
@@ -600,7 +591,25 @@ def test_fix_py3x_only_code(s, expected):
...
+
+@pytest.mark.xfail(reason='not implemented')
+def test_no_else_py3x_only_code():
+
+    s = (
+        'import sys\n'
+        'if sys.version_info >= (3, 6):\n'
+        '    pass'
+    )
+    expected = 'import sys\n'
```
asottile, line 614: I expect `import sys\npass\n` -- and my checkout of `pyupgrade` correctly makes that happen (?)

neutrinoceros: you're right, sorry, it's getting late in 🇫🇷

asottile (approved):

neutrinoceros:
wait don't merge I can add an id to the test

PR #534, opened 2021-09-15
Make `--py3X-plus` arguments mutually exclusive

Zac-HD:
I recently added similar arguments to `shed`, and thought you might appreciate a patch.


asottile:
they're intended to be last-one-wins so that you can chain them with scripting

Zac-HD:
Ah, makes sense 🙂 

(also please tell me if I'm opening too many PRs or issues; my current impression is that you're fine with just closing them like this if I'm proposing something out of scope)

asottile:
you're fine! thanks :D

Zac-HD:
np, just wanted to be sure 🤗

PR #552, opened 2021-10-05
Remove `from __future__ import annotations` on --py310-plus

### LINKED ISSUE #551: Remove `from __future__ import annotations` on `--py310-plus`?
Now that Python 3.10 is out, should pyupgrade remove `from __future__ import annotations`?

I can send a PR if that makes sense.

fsouza:
Also fixed the test name for generator_stop, not sure if that was
intended or I'm misreading it x)

Closes #551.


asottile:
https://github.com/asottile/pyupgrade/issues/551#issuecomment-934733789

Issue #549, opened 2021-09-30
Incorrect mock upgrade


vivster7:
I think I found a bug in the mock transforms.

`pyugrade --py3-plus` will change the transform the line:
`from mock import mock` -> `from unittest.mock import mock`

I believe the correct transform would be:
`from mock import mock` -> `from unittest import mock`

The `from unittest.mock import mock` line raises the following error:
```
  File "test.py", line 1, in <module>
    from unittest.mock import mock
ImportError: cannot import name 'mock'
```

I'd be happy to submit a patch for this if it will get accepted.

asottile:
the original import is incorrect, it should be `import mock`

vivster7:
Agreed that `import mock` should be preferred -- but `from mock import mock` does work as well without an import error. And since it works, people use it. https://grep.app/search?q=from%20mock%20import%20mock&case=true&words=true

asottile:
if you want to poke at the code it's here: https://github.com/asottile/pyupgrade/blob/master/pyupgrade/_plugins/mock.py

PR #554, opened 2021-10-11
fix relative import of mock module

vivster7:
Fixes https://github.com/asottile/pyupgrade/issues/549

## Description
`from mock import mock` is syntactically valid.
Without this patch, this transforms to `from unittest.mock import mock` (syntactically invalid)
With this patch, this line becomes `from unittest import mock`

## Details
This patch looks for the identifier `mock` in the `names` part of the `ImportFrom` AST node. Then, it handles this case in one of two ways:
 1. `len(names) == 1`.
   a. rewrite the module: `mock` -> `unittest`
 2. `len(names) > 1`. 
   a. rewrite the module: `mock` -> `unittest.mock`
   b. remove the identifier `mock` from the list of `names`.
   c. add import on a new line: `from unittest import mock`

### DIFF
--- a/pyupgrade/_plugins/mock.py
+++ b/pyupgrade/_plugins/mock.py
@@ -1,5 +1,6 @@
 import ast
-from typing import Iterable
+import functools
+from typing import Iterable, Optional
 from typing import List
 from typing import Tuple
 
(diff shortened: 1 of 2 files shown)


sadikkuzu:
@vivster7 
For avoiding those auto-fixes, you can run `pre-commit install` in your local folder.

asottile:
via #673

PR #562, opened 2021-10-24
replace exit(main()) with raise SystemExit(main())

asottile:
Committed via https://github.com/asottile/all-repos


WhyNotHugo:
What’s the advantage of raising vs calling exit?

asottile:
https://youtu.be/ZbeSPc5wL0g

WhyNotHugo:
Thanks, that was very clear!

PR #566, opened 2021-11-10
Rewrite docs examples with commented code to use diffs

adamchainz:
The diff format has been used for most new features and I find it much easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,38 +29,53 @@ Sample `.pre-commit-config.yaml`:
 
 ### Set literals
 
-```python
-set(())              # set()
-set([])              # set()
-set((1,))            # {1}
-set((1, 2))          # {1, 2}
-set([1, 2])          # {1, 2}
-set(x for x in y)    # {x for x in y}
-set([x for x in y])  # {x for x in y}
+```diff
+-set(())
++set()
+-set([])
++set()
+-set((1,))
++{1}
+-set((1, 2))
++{1, 2}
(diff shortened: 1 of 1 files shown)


asottile:
the commented approach is quite a bit more succinct when there's many examples or when they're multi-line examples

adamchainz:
This is true but I think that doesn't balance out the lack of readability. GitHub renders the comments faded out and italic by default, which makes comparison harder.

A table format could work for brevity. I'm also happy to remove the blank lines I've added here.

MarcoGorelli:
For what my opinion's worth, I think this is an improvement - the first time I stumbled upon this repo I couldn't tell what the comments were meant to be, it only made sense to me when I saw the diffs

hugovk:
https://github.com/asottile/setup-cfg-fmt#what-does-it-do also uses `diff` which looks good.

asottile:
yeah I guess removing the whitespace would be better?

adamchainz:
I removed the newlines that I addeed.

asottile (approved):

mxr:
@adamchainz any thoughts on this section that uses input/output instead of diffs?

https://github.com/asottile/pyupgrade/blob/ccf3a9192fca6d2883b1e67d785e0010243e30c3/README.md#python2-and-old-python3x-blocks

adamchainz:
It could look better as a diff, idk, try it out?

PR #580, opened 2021-12-27
rewrite abspath(__file__) to __file__ in py39+

### LINKED ISSUE #576: py39+: abspath(__file__) -> __file__
since `__file__` is now an absolute path in python3.9+

MarcoGorelli:
resolves #576

### DIFF
--- a/tests/features/abspath_file_test.py
+++ b/tests/features/abspath_file_test.py
@@ -1,0 +1,52 @@
+import pytest
+
+from pyupgrade._data import Settings
+from pyupgrade._main import _fix_plugins
+
+
+@pytest.mark.parametrize(
+    ('s', 'expected'),
+    (
+        pytest.param(
+            'from os.path import abspath\n'
+            'abspath(__file__)',
+            'from os.path import abspath\n'
+            '__file__',
+            id='abspath',
+        ),
+        pytest.param(
+            'import os\n'
+            'os.path.abspath(__file__)',
+            'import os\n'
+            '__file__',
+            id='os.path.abspath',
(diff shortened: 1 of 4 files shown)


tests/features/abspath_file_test.py
```diff
@@ -0,0 +1,52 @@
...
+        pytest.param(
+            'import os\n'
+            'os.path.abspath(__file__)',
+            'import os\n'
+            '__file__',
+            id='os.path.abspath',
+        ),
+    ),
+)
+def test_fix_encode(s, expected):
```
asottile, line 26: heh -- these got copy pasted

MarcoGorelli: caught me 😳

asottile (approved):

adamantike:
I will try to create a simple example and open a separate issue, but this change is breaking a script we are running with Python 3.9 and `alembic`.

We run `alembic upgrade --sql head`, while having a line like this in `alembic/env.py`:
```python
sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
```

That works, but after running `pyupgrade`, that line gets replaced with:
```python
sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```

And the `alembic` command fails. Printing `__file__` shows:
```python
ipdb> __file__
'alembic/env.py'
ipdb> import sys; sys.version
'3.9.13 (main, Aug  2 2022, 11:20:39) \n[GCC 10.2.1 20210110]'
```

Issue #415, opened 2021-03-30
pep 604 typing rewrites: incompatible with Pydantic objects.


sodul:
See https://github.com/samuelcolvin/pydantic/issues/2597 for details.

We use Python 3.9 with`from __future__ import annotations`. Pyupgrade will convert `Union[]` and `Optional[]` to the new pep 604 notation, which is very useful in the vast majority of our code but this conflicts with Pydantic which runs `eval()` on the types at runtime and where `pep 604` is not available, causing a `TypeError` during import time.

```
  File "/Users/stephane/my/module.py", line 45, in <module>
    class MyClass(pydantic.BaseModel):     # pylint: disable=no-member
  File "pydantic/main.py", line 247, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/typing.py", line 218, in pydantic.typing.resolve_annotations
    params = new_params
  File "/Users/stephane/.pyenv/versions/3.9.2/lib/python3.9/typing.py", line 283, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/Users/stephane/.pyenv/versions/3.9.2/lib/python3.9/typing.py", line 539, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

```py
from __future__ import annotations
import pydantic

class User(pydantic.BaseModel):     # pylint: disable=no-member
    first_name: str | None = None
    last_name: str | None = None
```

Would it be possible to disable the rewrite for Pydantic classes? At the moment we have to be careful about using Pyupgrade and undo typing rewrites on these classes.

Thank you!

asottile:
not to say RTFM but [RTFM](https://github.com/asottile/pyupgrade#pep-604-typing-rewrites)

sodul:
@asottile do you mean the `--keep-runtime-typing` option? It disables the new annotation for the entire file, not just the pydantic lines.

The request here is that this code:
```py
from __future__ import annotations
import pydantic

FOO: Optional[str] = None

class User(pydantic.BaseModel):
    first_name: Optional[str] = None
```

Would become:
```py
from __future__ import annotations
import pydantic

FOO: str | None = None

class User(pydantic.BaseModel):
    first_name: Optional[str] = None
```

[...]

PR #588, opened 2022-01-07
Improve readme

maresb:
I was really confused while trying to understand the README regarding #415. This PR would have cleared up my confusion.

1. I couldn't figure out whether the default boolean operation corresponding to a list is `and` or `or`. Turns out it's `or`.
2. There's no explanation of the purpose of `--keep-runtime-typing`.

Maybe I'm just exceptionally slow, but here was my train of thought:

I'm working on a project involving Pydantic which has been upgraded to 3.9. I know I need to disable PEP 604. But I don't want to disable PEP 585. The file imports `annotations` from `__future__`, so even if I use `--py39-plus` I'll lose PEP 585 (WRONG since 1. is not `and`!). Therefore, there is no flag which does what I want (WRONG because 2.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,7 +748,7 @@ Availability:
 
 Availability:
 - File imports `from __future__ import annotations`
-    - unless `--keep-runtime-typing` is passed on the commandline.
+    - unless `--keep-runtime-typing` is passed on the commandline (to avoid breaking Pydantic and similar tools).
 - or `--py39-plus` is passed on the commandline.
 
 ```diff
@@ -775,7 +775,7 @@ Availability:
 
 Availability:
 - File imports `from __future__ import annotations`
-    - unless `--keep-runtime-typing` is passed on the commandline.
+    - unless `--keep-runtime-typing` is passed on the commandline (to avoid breaking Pydantic and similar tools).
 - or `--py310-plus` is passed on the commandline.
 
 ```diff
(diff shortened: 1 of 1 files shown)


asottile:
hi -- while I appreciate the PR I don't think it needs clarification and I think the clarification makes it more difficult to read -- the same PR was previously rejected (though I can't find it since github search for  `or` isn't all that helpful)

addition of `or` shouldn't be necessary since the individual bullet points represent disparate cases (for example there's no need for `__future__.annotations` to enable it if the language version already supports it)

I also don't really want to advertise specific tools such as pydantic -- what they're doing is using runtime typing so the flag name is already appropriate for them

thanks for the PR nonetheless!

maresb:
As you wish. Of course it's clear to _you_ what you mean, since you're the one who wrote it! :laughing:

Even though I don't always agree, thanks for the extremely useful tools.

guillp:
Just for the record, I stumbled on exactly the same issue, which I noticed when pyupgrade did replace `Union[A, B]` with `A | B` in all my annotations even with `--py38-plus`. It is not obvious from official Python docs or from pyupgrade docs that the new syntax is valid on 3.8 when `from __future__ import annotations` is being used.

Some clarification would really help. The single word `or` at the beginning of some bullets points would remove a lot of confusion. Maybe you can reconsider this PR for inclusion?

Issue #162, opened 2019-06-11
Rewrite six.PY2 ternary


asottile:
```python
2 if six.PY2 else 3
```

this one is going to be quite tricky I imagine

graingert:
I'd probably implement this as two steps, one to replace `six.PY2` with `False` and another to remove unreachable branches

graingert:
@asottile here's a tool that does automatic dead code removal: https://github.com/coala/coala-bears/pull/2000

asottile:
@graingert note that this is already done for `if` statements, just not for ternary expressions -- no need to do two passes

`vulture` doesn't help here -- that dead code detection is similar to [dead](https://github.com/asottile/dead) and doesn't at all help us to remove code

the detection part is simple, the "finding the end and beginning of the ternary expression" is hard

graingert:
it seems they've managed to get it to apply removals to dead code detections https://github.com/jendrikseipp/vulture/issues/25

asottile:
cool, but it doesn't help us here

atugushev:
Hm... The trickiest part is to rewrite:

```python
3 if six.PY3 else 2
```

to

```python
3
```

asottile:
I think they're both nearly impossible given we don't properly know expression boundaries (though maybe the new python3.8 stuff helps?)

pope1ni:
I wonder if [PEP 657](https://www.python.org/dev/peps/pep-0657/#motivation) might help with this?

(It makes reference to branch differentiation in ternary statements w.r.t. coverage.)

asottile:
it does not help, we are not running code

pope1ni:
> it does not help, we are not running code

Of course. 🤦🏻‍♂️ Sorry for the noise. 🙂

juan-cloud95:
> ```python
> 2 if six.PY2 else 3
> ```
> 
> this one is going to be quite tricky I imagine

PR #603, opened 2022-02-16
Add rewriting of if expressions

aneeshusa:
Only handle the case where we keep the `else` branch,
as that provides an ending-delimiter before which we remove all tokens.

Part of https://github.com/asottile/pyupgrade/issues/162.


aneeshusa:
This duplicates a large chunk of code, let me know if/how you would want that deduped.
Also let me know if this half of https://github.com/asottile/pyupgrade/issues/162 seems safe to autofix or if there are edge cases I'm not thinking of.

aneeshusa:
Happy to add some more test coverage if this approach looks good to the maintainers.

Issue #606, opened 2022-02-21
Fix import of ABCs from 'collections' to come from 'collections.abc'


browniebroke:
Not sure if this was discussed before, the only discussions I could find mentioned the ones coming from `typing`, which seems different to me.

I was looking at some warning raised in one of my projects during tests and this one came up:

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Generator
```

I looked at the previous issues, and all I could find are #522 and #558 which AFAIU are different, they suggest to modify the `typing` import. The change I had to make to fix this warning is the following:

```diff
- from collections import Generator
+ from collections.abc import Generator
```

We're using pyupgrade [v2.29.1](https://github.com/asottile/pyupgrade/releases/tag/v2.29.1), which doesn't seem to handle this, and looking at the changes for more recent releases, it doesn't look like it was added recently. Is there any reason I'm missing why it couldn't be handled by pyupgrade? I'm happy to contribute a pull request if it's a welcome addition. Here is the list of all the affected imports I found (using Python 3.8):

- Hashable
- Awaitable
- Coroutine
- AsyncIterable
- AsyncIterator
- Iterable
- Iterator
- Reversible
- Sized
- Container
- Collection
- MutableSet
- Mapping
- MutableMapping
- Sequence
- MutableSequence
- ByteString
- MappingView
- KeysView
- ItemsView
- ValuesView
- Generator
- AsyncGenerator

How to reproduce:

```bash
> python -X dev -c 'from collections import Generator'
<string>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
```

PS: thanks for this amazing package

ptmcg:
We worked on this issue during our hackathon at work this week, including handling of the imports of mixed ABC's and non-ABC's - PR forthcoming.  Processed 500,000 LOC, found and corrected 34 instances (@browniebroke's test cases look great, I will add them with our own).

asottile:
this has been handled by #661 #663 #664 #665 #666 #667 #668 #669 #670 #671 #672

PR #607, opened 2022-02-21
rewrite collections imports for ABC classes to collections.abc

browniebroke:
Ref #606 

- [x] rewrite imports when all imported classes are ABC
- [x] document the feature in the README

I'm tempted to leave Ithe 2 more complex cases out for now:

- [ ] support mix of ABC and non-ABC imported classes?
- [ ] support absolute import?

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -632,6 +632,60 @@ Availability:
 +float
 ```
 
+### Rewrite imports of ABC from `collections`
+
+Availability:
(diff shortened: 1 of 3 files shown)


cclauss (approved):

asottile:
these imports are now rewritten on the main branch -- I validated your testcases and even the "partial" import one passes now as well!

thanks again o7

PR #629, opened 2022-04-09
Added plugins for changing imports and references to ABC's from colle…

ptmcg:
…ctions module to collections.abc module (Issue #606)


asottile:
this has been handled by #661 #663 #664 #665 #666 #667 #668 #669 #670 #671 #672

PR #638, opened 2022-05-04
Do not remove quoted annotations on 3.11

### LINKED ISSUE #637: `--py311-plus` should not unquote annotations
https://github.com/asottile/pyupgrade#remove-quoted-annotations says that `--py311-plus` will remove quoted annotations. I suppose that was done under the assumption that PEP 563 (`from __future__ import annotations`) would become the default in 3.11, but that's not happening. (And probably not in the future either.)

JelleZijlstra:
Fixes #637

### DIFF
--- a/README.md
+++ b/README.md
@@ -780,7 +780,6 @@ Availability:
 
 Availability:
 - File imports `from __future__ import annotations`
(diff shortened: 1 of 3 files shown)


asottile (approved):
tyty

PR #643, opened 2022-05-15
Add Python version setting example for config

ulgens:
I thought it's better to have the example target Python version setting in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Sample `.pre-commit-config.yaml`:
     rev: v2.32.1
     hooks:
     -   id: pyupgrade
+        args: [--py310-plus]  # Optional
 ```
 
 ## Implemented features


asottile:
there's no reason to duplicate [pre-commit's](https://pre-commit.com) docs poorly here

ulgens:
@asottile What do you mean? `--py310-plus` or other Python version specifier's are not pre-commit's parameters. Python version specifier is largely used with pyupgrade in my experience, but still it's not mentioned in readme which I think something needs to be fixed.

asottile:
pre-commit shows you [how to use `args`](https://pre-commit.com/#passing-arguments-to-hooks), the readme mentions the current set of options per feature, and as with any decent commandline `--help` works

ulgens:
And I'm telling that instead of making the user discover those parameters themselves via CLI, it's a much a better experience to put them on the readme but I see we are not on the same page. Thanks for your time.

asottile:
yes we are not on the same page, are you consulting the same README?  it shows `--py310-plus` right here for example: https://github.com/asottile/pyupgrade#pep-604-typing-rewrites

ulgens:
I am consulting the same README, yes. Not sure why you are so aggressive with your tone and I'm willing to ignore it, but for a new user of the tool, which I had to deal with tens of times because I'm pushing other people to use it, a simple example about that parameter is much more useful than mentioning it an irrelevant place in a different context. 

Thanks again.

asottile:
you seem to be misunderstanding -- not sure where you're picking up tone or aggression but sure have a good one

Issue #631, opened 2022-04-13
Feature Proposal: Rewrite `unittest.makeSuite|getTestCaseNames|findTestCases`.


sallner:
With the patch [bpo-5846](https://github.com/python/cpython/pull/28382) three methods from the `unittest` module have been deprecated. It can be substituted by `unittest.defaultTestLoader.loadTestsFromTestCase|getTestCaseNames|findTestCases`. (https://github.com/zopefoundation/zope.security/pull/84)

Would it be an option to include this fix in `pyupgrade` and would you accept such a fixer? This would help a lot with the whole range of `zope.*` packages.

asottile:
there's already a unit test rewriter, it would just need to learn about these functions, would you like to contribute a patch?

sallner:
Thank you for the hint. I can try that.

cj81499:
Hey @asottile and @sallner. I took a shot at implementing this. Take a look when you get a chance :)

PR #647, opened 2022-06-04
feat: rewrite deprecated unittest methods

cj81499:
resolves https://github.com/asottile/pyupgrade/issues/631


pyupgrade/_plugins/unittest_aliases.py
```diff
@@ -12,50 +12,67 @@
...
 
 @register(ast.Call)
 def visit_Call(
-        state: State,
-        node: ast.Call,
-        parent: ast.AST,
+    state: State,
+    node: ast.Call,
+    parent: ast.AST,
 ) -> Iterable[tuple[Offset, TokenFunc]]:
```
asottile, line 55: generally shouldn't touch formatting unrelated to your change -- also I prefer these double-indented to differentiate from the code

asottile (commented):
this PR tries too hard to _DRY_ -- I would write the `unittest.*` ones separately (becomes a ~4 line patch rather than a 60 line patch)

pyupgrade/_plugins/unittest_aliases.py
cj81499: I must've accidentally ran `black`. Will revert.

cj81499:
> ~4 line patch

I'm not sure I understand how this patch could ever be that small. I figure that at a minimum, this needs:

- 3 lines for the new replacement rules
- updates to 2 lines with hardcoded `"self"`

Regardless, I'm happy un-DRY it. Are you looking for these replacements to be an entirely different plugin? Or an additional `dict` (for `unittest` prefixed substitutions)? Or something else?

asottile:
> > ~4 line patch
> 
> I'm not sure I understand how this patch could ever be that small. I figure that at a minimum, this needs:
> 
>     * 3 lines for the new replacement rules
> 
>     * updates to 2 lines with hardcoded `"self"`
> 
> 
> Regardless, I'm happy un-DRY it. Are you looking for these replacements to be an entirely different plugin? Or an additional `dict` (for `unittest` prefixed substitutions)? Or something else?

~essentially leaving everything as-is but these 4 additions:

1 new dictionary
1 new elif statement below the existing one
1 `partial(...)` call
1 yield statement below that

cj81499:
You don't happen to have a formatter that'll enforce the (imo, weird) indentation rules you're using, do you? I keep messing it up accidentally 😆

I'm really used to just letting `black` run on save and forgetting about it entirely.

pyupgrade/_plugins/unittest_aliases.py
```diff
@@ -59,3 +64,15 @@ def visit_Call(
...
+            isinstance(node.func.value, ast.Name) and
+            node.func.value.id == 'unittest' and
+            node.func.attr in FUNCTION_MAPPING
+    ):
+        func = functools.partial(
+            replace_name,
+            name=node.func.attr,
+            new=f'unittest.{FUNCTION_MAPPING[node.func.attr]}',
+        )
+        yield ast_to_offset(node.func), func
```
asottile, line 78: ahah, I'm glad we did this -- now that the code isn't muddled with DRYness it makes the subtle bug way more obvious:

these attributes are only available in a specific version of python -- right now this is performing the rewrite on all target versions

asottile:
someone in my chat is working on a formatter, though I don't know if it's done yet

as for squash I'd prefer you'd do it because I always make merge commits but if you don't I'll "good code changed like a ghost" it for you

pyupgrade/_plugins/unittest_aliases.py
cj81499: > these attributes are only available in a specific version of python -- right now this is performing the rewrite on all target versions

Believe it or not, I thought about that and checked ahead of time! Everything we need seem to exist in 2.7+:

[defaultTestLoader](https://docs.python.org/2.7/library/unittest.html?highlight=defaulttestloader#unittest.defaultTestLoader)

[loadTestsFromModule](https://docs.python.org/2.7/library/unittest.html?highlight=defaulttestloader#unittest.TestLoader.loadTestsFromModule)

[loadTestsFromTestCase](https://docs.python.org/2.7/library/unittest.html?highlight=defaulttestloader#unittest.TestLoader.loadTestsFromTestCase)

[getTestCaseNames](https://docs.python.org/2.7/library/unittest.html?highlight=defaulttestloader#unittest.TestLoader.getTestCaseNames)

cj81499:
> someone in my chat is working on a formatter, though I don't know if it's done yet

Nice!

> as for squash I'd prefer you'd do it because I always make merge commits but if you don't I'll "good code changed like a ghost" it for you

On it!

pyupgrade/_plugins/unittest_aliases.py
```diff
@@ -12,7 +12,6 @@
 from pyupgrade._data import TokenFunc
 from pyupgrade._token_helpers import replace_name
 
-
```
asottile, line 15: oops, one little thing -- or I can fix it -- up to you

cj81499: resolved

asottile (approved):

PR #650, opened 2022-06-06
rewrite bytes(b'foo') to b'foo'

### LINKED ISSUE #649: rewrite bytes(b'foo') -> b'foo'
looks like this was a somewhat common mistake when doing a python2/python3 port

similar to the `str('string literal')` fix that's already happening in pyupgrade

theendlessriver13:
resolves #649

I unfortunatelly have to check `node.func.id == 'bytes'` twice, to later narrow it down so that the Call name is `bytes` **while** the arg is `ast.Bytes`. But for some reason mypy is mad at the second check with:
```
pyupgrade/_plugins/native_literals.py:56: error: "expr" has no attribute "id"
Found 1 error in 1 file (checked 95 source files)
```
could you give me a hint what's wrong there?

also not sure if the rename of the `_fix_native_str` function is okay since it now also handles bytes?

### DIFF
--- a/pyupgrade/_plugins/native_literals.py
+++ b/pyupgrade/_plugins/native_literals.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import ast
+import functools
 from typing import Iterable
 
 from tokenize_rt import Offset
@@ -19,7 +20,7 @@ from pyupgrade._token_helpers import replace_call
 SIX_NATIVE_STR = frozenset(('ensure_str', 'ensure_text', 'text_type'))
 
 
-def _fix_native_str(i: int, tokens: list[Token]) -> None:
+def _fix_literal(i: int, tokens: list[Token], *, empty: str) -> None:
     j = find_open_paren(tokens, i)
     func_args, end = parse_call_args(tokens, j)
     if any(tok.name == 'NL' for tok in tokens[i:end]):
@@ -27,7 +28,7 @@ def _fix_native_str(i: int, tokens: list[Token]) -> None:
     if func_args:
         replace_call(tokens, i, end, func_args, '{args[0]}')
     else:
-        tokens[i:end] = [tokens[i]._replace(name='STRING', src="''")]
+        tokens[i:end] = [tokens[i]._replace(name='STRING', src=empty)]
 
 
 def is_a_native_literal_call(
@@ -58,4 +59,16 @@ def visit_Call(
             state.settings.min_version >= (3,) and
             is_a_native_literal_call(node, state.from_imports)
     ):
-        yield ast_to_offset(node), _fix_native_str
+        func = functools.partial(_fix_literal, empty="''")
+        yield ast_to_offset(node), func
+    elif (
+            state.settings.min_version >= (3,) and
+            isinstance(node.func, ast.Name) and node.func.id == 'bytes' and
+            not node.keywords and not has_starargs(node) and
+            (
+                len(node.args) == 0 or
+                (len(node.args) == 1 and isinstance(node.args[0], ast.Bytes))
+            )
+    ):
+        func = functools.partial(_fix_literal, empty="b''")
+        yield ast_to_offset(node), func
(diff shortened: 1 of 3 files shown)


pyupgrade/_plugins/native_literals.py
```diff
@@ -58,4 +71,4 @@ def visit_Call(
             state.settings.min_version >= (3,) and
             is_a_native_literal_call(node, state.from_imports)
```
asottile, line 72: this function name and implementation no longer matches -- I would do a separate condition entirely for bytes

theendlessriver13:
for the coverage issue:
I could either make it just `else` instead of the `elif` checking for `bytes` or add another `else` at the end with raising `NotImplementedError`.

I think the former with just else is fine, since it __should__ only ever be `str` or `bytes`.

pyupgrade/_plugins/native_literals.py
theendlessriver13: I added a `is_a_bytes_from_bytestr_call` check. I feel like splitting up the `_fix_native_*` function does not make too much sense, since both would only be different by one character (`src="b''"`) or is adding the `if` a big overhead?

pyupgrade/_plugins/native_literals.py
```diff
@@ -19,15 +19,33 @@
 SIX_NATIVE_STR = frozenset(('ensure_str', 'ensure_text', 'text_type'))
 
 
-def _fix_native_str(i: int, tokens: list[Token]) -> None:
+def _fix_native_literal(i: int, tokens: list[Token]) -> None:
```
asottile, line 22: ah ok, yeah the confusion here -- "native" refers to that `str` is the type of `"a string"` independent of python2 / python 3

this function would probably be simpler by having a parameter which is the intended "empty" value -- rather than having to double-check and add if statements

pyupgrade/_plugins/native_literals.py
```diff
@@ -56,6 +74,9 @@ def visit_Call(
...
     if (
             state.settings.min_version >= (3,) and
-            is_a_native_literal_call(node, state.from_imports)
+            (
+                is_a_native_literal_call(node, state.from_imports) or
+                is_a_bytes_from_bytestr_call(node, state.from_imports)
+            )
     ):
-        yield ast_to_offset(node), _fix_native_str
+        yield ast_to_offset(node), _fix_native_literal
```
asottile, line 82: I would add a separate `elif` for the bytestring check -- especially given the suggestion above

pyupgrade/_plugins/native_literals.py
```diff
@@ -19,15 +19,33 @@
...
+) -> bool:
+    return (
+        isinstance(node.func, ast.Name) and node.func.id == 'bytes' and
+        not node.keywords and
+        not has_starargs(node) and
+        (
+            len(node.args) == 0 or
+            (len(node.args) == 1 and isinstance(node.args[0], ast.Bytes))
+        )
+    )
```
asottile, line 48: the other function was written separately because it was reused in the `six_simple` plugin -- this one doesn't need to be a function

pyupgrade/_plugins/native_literals.py
theendlessriver13: sorry, I don't fully understand this yet, so you mean something like:
```python
def _fix_something_something(i: int, tokens: list[Token], *, empty: str | bytes) -> None:
    ...
    else:
        tokens[i:end] = [tokens[i]._replace(name='STRING', src=empty)]
```
but how can we pass this along when yielding the `TokenFunc` at the end?

asottile: `git grep` around for `functools.partial` -- there's a bunch of examples

it wouldn't be `str | bytes` -- just `str` (it's source)

pyupgrade/_plugins/native_literals.py
```diff
@@ -19,15 +20,15 @@
 SIX_NATIVE_STR = frozenset(('ensure_str', 'ensure_text', 'text_type'))
 
 
-def _fix_native_str(i: int, tokens: list[Token]) -> None:
+def _fix_native_literal(i: int, tokens: list[Token], *, empty: str) -> None:
```
asottile, line 23: sorry!  one more thing, let's just call this `_fix_literal` since it has nothing to do with `native` now

asottile (approved):

Issue #273, opened 2020-03-19
two steps to converge: io.open(..., 'r')


asottile:
the first step goes to `open(..., 'r')` and the second step goes to `open(...)`

PR #655, opened 2022-06-25
io.open two steps to converge through open

hamdanal:
Hey Anthony, would something like this work for #273 ?


pyupgrade/_plugins/io_open.py
```diff
@@ -32,3 +33,8 @@ def visit_Call(
             node.func.attr == 'open'
     ):
         yield ast_to_offset(node.func), _replace_io_open
+
+        # apply open() upgrades
+        node.func = node.func.value
+        node.func.id = 'open'
```
asottile, line 39: the ast should be treated as immutable -- this is a hack I can't accept

hamdanal: Faire enough.
If you have any idea/suggestion/guidance on how to implement it without the hack, I will happily give it another try, otherwise feel free to close the PR.

asottile: my guess is that the `open` check would just need to learn about `io.open` and also probably the `io.open` rewriter would need to keep the same offsets when rewriting (if it doesn't already)

asottile:
hi!  thanks for the PR -- I took the idea that I outlined [here](https://github.com/asottile/pyupgrade/pull/655#discussion_r906689732) and ended up implementing that in #656 -- I added you [as credited](https://youtu.be/_-qNX8EbVYI) on the commit -- sorry it didn't work out exactly in this PR but thank you nonetheless for the effort on this!

hamdanal:
No worries, thanks for the credits. Have a good one.

PR #686, opened 2022-07-26
fix imports removal not including rules from previous versions

Senpos:
reported by `bork` in [discord](https://discord.com/channels/576802746850869258/747133137804591156/1001211167114530856)

when using `--py38-plus` (or any other mode which is a superset of `py3-plus`), most of the imports were not removed
```python
$ cat t.py
from __future__ import print_function

pass
$ pyupgrade --py38-plus t.py
$ cat t.py
from __future__ import print_function

pass
$ pyupgrade --py3-plus t.py
Rewriting t.py
$ cat t.py
pass
```
introduced by refactoring in https://github.com/asottile/pyupgrade/commit/4775ca8ad7cac72dc8dd5c6bcd29defd2d375cbb

added a test-case

### DIFF
--- a/pyupgrade/_plugins/imports.py
+++ b/pyupgrade/_plugins/imports.py
@@ -233,10 +233,11 @@ def _for_version(
         Mapping[tuple[str, str], str],
         Mapping[str, str],
 ]:
-    removals = {}
+    removals = collections.defaultdict(set)
     for ver, ver_removals in REMOVALS.items():
         if ver <= version:
-            removals.update(ver_removals)
+            for base, names in ver_removals.items():
+                removals[base].update(names)
 
     exact = {}
     for ver, ver_exact in REPLACE_EXACT.items():
--- a/tests/features/import_removals_test.py
+++ b/tests/features/import_removals_test.py
@@ -39,6 +39,7 @@ def test_import_removals_noop(s, min_version):
         ('from __future__ import division\n', (3,), ''),
         ('from __future__ import division\n', (3, 6), ''),
         ('from __future__ import (generators,)', (2, 7), ''),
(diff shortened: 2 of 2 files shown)


pyupgrade/_plugins/imports.py
```diff
@@ -233,10 +233,11 @@ def _for_version(
         Mapping[tuple[str, str], str],
         Mapping[str, str],
 ]:
-    removals = {}
+    removals: dict[str, set[str]] = {}
     for ver, ver_removals in REMOVALS.items():
         if ver <= version:
-            removals.update(ver_removals)
+            for base, names in ver_removals.items():
+                removals.setdefault(base, set()).update(names)
```
asottile, line 240: `collections.defaultdict(set)` would be better

tests/features/import_removals_test.py
```diff
@@ -127,6 +127,7 @@ def test_import_removals_noop(s, min_version):
             'from io import BytesIO as BIO\n',
             id='removal with import-as',
         ),
+        ('from __future__ import print_function', (3, 8), ''),
```
asottile, line 130: this should be up by the other ones

asottile:
the commit message is incorrect and doesn't explain the fix

pyupgrade/_plugins/imports.py
Senpos: Done.
Initially I decided to avoid it, so it won't introduce any surprises when accessing missing keys and if there was any logic aronud it, but it seems there should be no issues with that.

tests/features/import_removals_test.py
Senpos: Done!

asottile (approved):

PR #699, opened 2022-08-29
Add "Only diff" section to README.md

dnoliver:
Took me a while to figure out those commands. 
They allow me to run this task in a CI environment without having to install pre-commit and write a yaml file.
Not actually expecting this to be merged, but just as a way of documenting and alternative way of running this is a CI environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,16 @@ Sample `.pre-commit-config.yaml`:
     -   id: pyupgrade
 ```
 
+## Only diff
+
+The following command will exit with non-zero exit code if there are changes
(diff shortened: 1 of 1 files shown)


asottile:
creative -- though I don't think documenting how to use `diff` has a place in `pyupgrade`'s readme -- thanks for the PR nonetheless!

Viicos:
What about a `--diff` flag similar to `black` for instance? I think there's a lot of use cases where we'd like to only see the diff being outputed, and keep files untouched

asottile:
no means no. please respect

Issue #697, opened 2022-08-19
Preserve comments when rewriting NamedTuples to class style


PeterJCLaw:
``` diff
 from typing import NamedTuple
 
-Thing = NamedTuple('Thing', (
-    ('value', int),  # comment
-))
+class Thing(NamedTuple):
+    value: int
```
Ideally the comment should be preserved.

asottile:
this isn't feasible -- for instance what should it do for

```python
Thing = NamedTuple('Thing', (  # comment1
    # comment2
    ('value', int),  # comment3
    # comment4
))  # comment5
```

asottile:
also please search the issue tracker next time, there are several duplicates

PeterJCLaw:
> this isn't feasible -- for instance what should it do for
> 
> ```python
> Thing = NamedTuple('Thing', (  # comment1
>     # comment2
>     ('value', int),  # comment3
>     # comment4
> ))  # comment5
> ```

I suggest either rejecting the block or doing this:

```python
class Thing(NamedTuple):  # comment1
    # comment2
    value: int  # comment3
    # comment4
  # comment5
```

This may not be the prettiest, but avoids throwing away information. Yes, most users _probably_ are using version control, however I suspect some aren't. Even those using version control are likely to miss this sort of thing if making changes to a non-trivial codebase -- I don't particularity fancy reviewing automated changes to projects over about 1k LOC, do you?

> also please search the issue tracker next time, there are several duplicates

Apologies, however does the repeated reporting of this not suggest it's worth doing something about? Even just rejecting the block would be preferable IMO.

Issue #634, opened 2022-05-01
pyupgrade removes comments from NamedTuple


hauntsaninja:
Thanks for pyugprade, it's great!

```
from typing import NamedTuple, List

CacheMeta = NamedTuple('CacheMeta',
                       [('id', str),
                        ('dependencies', List[str]),  # names of imported modules
                        ('data_mtime', int),  # mtime of data_json
                        ('data_json', str),  # path of <id>.data.json
])
```

Running `pyupgrade --py36-plus test.py` gives:
```
from typing import NamedTuple, List

class CacheMeta(NamedTuple):
    id: str
    dependencies: List[str]
    data_mtime: int
    data_json: str
```

hauntsaninja:
Oops, didn't search closed issues, duplicate of https://github.com/asottile/pyupgrade/issues/497

Issue #497, opened 2021-07-22
pyupgrade can eat comments on typing.NamedTuple


ethanhs:
With the given file:

```python
# test.py
from typing import NamedTuple, Optional

class TypeInfo:
    pass

DeferredNode = NamedTuple(
    'DeferredNode',
     [
        ('active_typeinfo', Optional[TypeInfo]),  # And its TypeInfo (for semantic analysis
                                                  # self type handling)
    ])
```

run:
```
$ pyupgrade --py36-plus test.py
```

And the result is:
```python
from typing import NamedTuple, Optional

class TypeInfo:
    pass

class DeferredNode(NamedTuple):
    active_typeinfo: Optional[TypeInfo]
```

EDIT: I should note this was found when running pyupgrade on mypy.

ethanhs:
Okay this is bizarre - apparently if you delete the assignment *target* the comments don't get deleted... wat

asottile:
yeah this isn't really possible to preserve -- the entire call to `NamedTuple` is removed and replaced

this is (sadly) one of the places where preserving comments isn't possible (another is some of the more advanced multi-line six calls)

ethanhs:
Ah ok, didn't know!

PR #700, opened 2022-09-08
Implement preservation of comments "within" typing classes

PeterJCLaw:
This ensures that comments which appear around `TypedDict`s and `NamedTuple`s that are converted to class style are preserved.

The layout of the comments in the new code isn't perfect, however that's relatively easy for the user to fix manually afterwards. This is the main negative of this approach. There are things which we could do here (such as looking at the previous/next tokens and comments' tokens to try to line them up), however IMO having the comments with slight alignment issues is useful enough even with this limitation.

I've added tests based on the cases I'm aware of, but very open to adding more tests if there are cases you're concerned about which are not covered here.

Fixes https://github.com/asottile/pyupgrade/issues/497
Fixes https://github.com/asottile/pyupgrade/issues/634
Fixes https://github.com/asottile/pyupgrade/issues/697

### DIFF
--- a/pyupgrade/_plugins/typing_classes.py
+++ b/pyupgrade/_plugins/typing_classes.py
@@ -66,10 +66,30 @@ def _typed_class_replacement(
     # NT = NamedTuple("nt", [("a", int)])
     # ^i                                 ^end
     end = i + 1
+    comments = []
     while end < len(tokens) and tokens[end].name != 'NEWLINE':
+        token = tokens[end]
+        if token.name == 'COMMENT':
+            comments.append(token)
         end += 1
 
-    attrs = '\n'.join(f'{indent}{k}: {_unparse(v)}' for k, v in types.items())
+    parts = []
+    for k, v in types.items():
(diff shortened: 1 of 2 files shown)


pyupgrade/_plugins/typing_classes.py
```diff
@@ -66,10 +66,30 @@ def _typed_class_replacement(
...
+    for k, v in types.items():
+        while comments and (v.lineno, v.col_offset) > comments[0].offset:
+            comment = comments.pop(0)
+            parts.append(f'{indent}{comment.src}')
+
+        member = f'{indent}{k}: {_unparse(v)}'
+
+        if comments and v.lineno == comments[0].line:
+            comment = comments.pop(0)
+            member += f' {comment.src}'
```
asottile, line 86: this is going to immediately trigger E261

PeterJCLaw: Ah, sorry. 4dcc460d4642afb91c3e42232e8561e2eef6337c should fix that.

asottile (approved):

Issue #543, opened 2021-09-24
Revert converting str.join(listcomp) to str.join(generator)


sethmlarson:
Consider reverting the listcomp-generator for the `str.join()` case, it's much faster to use a list comprehension than a generator in this case:

```bash
$ python -m timeit '"y".join("x" for _ in range(1000000))'
5 loops, best of 5: 40.6 msec per loop
$ python -m timeit '"y".join(["x" for _ in range(1000000)])'
10 loops, best of 5: 29.6 msec per loop
```

This is because `str.join()` constructs a list internally and short-circuits if passed an already created list.

MarcoGorelli:
Hasn't this already been reverted in https://github.com/asottile/pyupgrade/issues/456 ? Which version are you using?

sethmlarson:
@MarcoGorelli I believe we're using v2.27.0, this pull request is where the diff was created: https://github.com/urllib3/urllib3/pull/2431

asottile:
you're on an old version

MarcoGorelli:
There's no `join` in that PR, as far as I can tell - what there is is a comment about not doing the list comp -> generator comp for `frozenset`. Is that what you meant?

sethmlarson:
@MarcoGorelli Thanks for making my eyes actually work, sorry for the noise.

MarcoGorelli:
Having tried this out, it looks like even that one's slower:
```
In [8]: %timeit frozenset([i for i in range(1000)])
36.6 µs ± 661 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [9]: %timeit frozenset((i for i in range(1000)))
41.8 µs ± 940 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

I'd suggested this feature, and it seems that in several cases it induces a small performance hit, so...I'm sorry about that

@asottile reckon it's worth removing this rewrite altogether?

PR #705, opened 2022-09-12
remove pep289 rewrite

### LINKED ISSUE #704: pyupgrade optimization issue: tuple + list comprehension
pyupgrade transform this line:

```python
tuple([x for x in my_range])
```
in the less efficient:

```python
tuple(x for x in my_range)
```

Have a look at this example:

```console
[ale@flo pyupgrade-issue]$ cat a.py 
my_range = range(10000)

for i in my_range:
    tuple([x for x in my_range])
```

This script runs in less than 2 seconds:

```console
[ale@flo pyupgrade-issue]$ time python3.10 a.py 

real    0m1,812s
user    0m1,734s
sys     0m0,076s
```

Buth after I run pyupgrade I see it gets significantly slower:

```console
[ale@flo pyupgrade-issue]$ pyupgrade a.py 
WARNING: pyupgrade will default to --py3-plus in 3.x
Rewriting a.py
[ale@flo pyupgrade-issue]$ git diff
...
-    tuple([x for x in my_range])
+    tuple(x for x in my_range)
[ale@flo pyupgrade-issue]$ time python3.10 a.py 

real    0m2,403s
user    0m2,394s
sys     0m0,009s
```

MarcoGorelli:
closes #704 
xref https://github.com/asottile/pyupgrade/issues/543

Apologies for having introduced this rewrite - from the recent issue, and other issues, I think it's done more harm than good, so am suggesting to just remove it altogether


asottile:
it's a shame cpython doesn't use the generator length hints to make these fast

eli-schwartz:
Wouldn't this transform still be good to have, for the case of `any()` and `all()`? Which I notice was not originally on the list of ALLOWED_FUNCS...

I'm not well aware of how/why the performance suffers for any given function (pep 289 rationalizes this for memory, not speed particularly), but... `any()`/`all()` are special cases where the function's debatably primary use is to get a generator, and short-circuit that generator under various conditions, so it should be safe.

PR #707, opened 2022-09-17
Rewrite 2-arg super call in nested class

### LINKED ISSUE #696: super() calls in nested classes aren't updated
It appears that calls to `super()` within inner/nested classes aren't updated as I would expect.

Minimal reproduction:

`.pre-commit-config.yaml`:
```yaml
repos:
-   repo: https://github.com/asottile/pyupgrade
    rev: v2.37.3
    hooks:
    -   id: pyupgrade
        args: [--py3-plus]
```

`test.py`:
```python
class A:
    def __init__(self):
        super(A, self).__init__()

    class B:
        def __init__(self):
            super(A.B, self).__init__()
```

```diff
>git init
>git add test.py
>pre-commit run
pyupgrade................................................................Failed
- hook id: pyupgrade
- exit code: 1
- files were modified by this hook

Rewriting test.py

>git diff -U10
diff --git a/test.py b/test.py
index a58b7cd..abdff2e 100644
--- a/test.py
+++ b/test.py
@@ -1,7 +1,7 @@
 class A:
     def __init__(self):
-        super(A, self).__init__()
+        super().__init__()

     class B:
         def __init__(self):
             super(A.B, self).__init__()
```

I would have expected to instead see:
```diff
diff --git a/test.py b/test.py
index a58b7cd..b2e4b1b 100644
--- a/test.py
+++ b/test.py
@@ -1,7 +1,7 @@
 class A:
     def __init__(self):
-        super(A, self).__init__()
+        super().__init__()

     class B:
         def __init__(self):
-            super(A.B, self).__init__()
+            super().__init__()
```

There is a [test case](https://github.com/asottile/pyupgrade/blob/ad5d9db9a206bfd221760fd81e407bf6040c808c/tests/features/super_test.py#L86-L95) for nested classes, but I don't believe it is valid.

In this test case, the reference to `C` in the super call will lead to `NameError: name 'C' is not defined`. It instead needs to be `Outer.C`.
```python
class Outer:
    class C(Base):
        def f(self):
            super(C, self).f()
```

hamdanal:
Resolves #696

### DIFF
--- a/tests/features/super_test.py
+++ b/tests/features/super_test.py
@@ -20,6 +20,19 @@ from pyupgrade._main import Settings
         'class C(Base):\n'
         '    def f(self):\n'
         '        super(Base, self).f()\n',
+        'class Outer:\n'  # common nesting
+        '    class C(Base):\n'
+        '        def f(self):\n'
+        '            super(C, self).f()\n',
+        'class Outer:\n'  # higher levels of nesting
+        '    class Inner:\n'
+        '        class C(Base):\n'
+        '            def f(self):\n'
+        '                super(Inner.C, self).f()\n',
+        'class Outer:\n'  # super arg1 nested in unrelated name
+        '    class C(Base):\n'
+        '        def f(self):\n'
+        '            super(some_module.Outer.C, self).f()\n',
 
         # super outside of a class (technically legal!)
         'def f(self):\n'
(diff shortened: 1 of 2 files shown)


asottile:
one of the existing tests is wrong so I expect some change there too

hamdanal:
fixed

tests/features/super_test.py
```diff
@@ -87,7 +91,7 @@ def test_fix_super_noop(s):
             'class Outer:\n'
             '    class C(Base):\n'
             '        def f(self):\n'
-            '            super (C, self).f()\n',
+            '            super(Outer.C, self).f()\n',
```
asottile, line 94: while the space looks weird, it's intentional to test an edge case

asottile (commented):
this should generalize -- I don't think that's happening at the moment:

```python
class A:
    class B:
        class C:
            def f(self):
                super(A.B.C, self).f()
```

asottile:
you should add the example I put in my comment as a test case

hamdanal:
I can squash the commits if you'd like to keep the commit history clean

asottile:
> I can squash the commits if you'd like to keep the commit history clean

I simplified the logic a bit too -- counting down vs. counting up is a lot easier to manage

asottile (approved):

PR #717, opened 2022-10-04
Add flags for keep-open-mode and keep-format-indices

scottkmaxwell:
Some of our users will prefer to be explicit. How do you feel about adding these switches?


asottile:
no sorry -- flags are only for breakages not for opinions

scottkmaxwell:
OK, that's fine. But I would argue that these changes are not "upgrades", but stylistic choices. The mode has been optional since at least Py2.4 and the auto-numbering in format has been in since Py2.7. Some of my developers prefer to be explicit. I cannot impose those stylistic choices on all of my development teams, so I guess I'll need to use my fork.

PR #718, opened 2022-10-04
remove python-2-compatible rewriting mode

asottile:


EwoutH:
You just closed #722, the reason I opened that is that a major release was tagged without any indication what's new in it by itself. All the information given in the git log and this PR is _remove python-2-compatible rewriting mode_. I suspect that means that Python files can now only be formatted in Python 3 format. But does it still allow Python 2 files to be formatted? Is suspect yes, but can't know for sure.

A changelog is a means of communication, as are commit messages and PR descriptions. Of course everyone could dive into the code changes and what changed. But they can't see why it changed. Or what the considerations were. Or what unexpected limitations or behaviour could occur. Or why it changed now.

For example, a lot of pre-commit CI configs will be updated to 3.0.0. Maintainers want to quickly see what's new in this major version, and what the caveats are. Some CI builds will break, and then they will want to see what mitigations are possible. For example, what happens if you input the `--py3-plus` parameter? Will v3.0.0 crash? And if so, probably removing `--py3-plus` helps. But does v3.0.0 provide the same functionality as `--py3-plus` did in v2.0? Those are legit questions many maintainers will ask and now need to individually investigate.

This is an amazing piece of tooling and a fantastic quality of life enhancer for the Python ecosystem. But good documentation is key for keeping it accessible and understandable. So please invest in that.

asottile:
please don't tell me how to run my projects thanks -- if you want to demand more of my free time then support the "amazing piece of tooling" and sponsor.

PR #721, opened 2022-10-04
Fix "wt" and keep "r" for open mode

scottkmaxwell:
This updates the redundant open mode in 2 ways:
- replaces `"wt"` and `"tw"` with `"w"` (was a TODO in the tests)
- retains `"r"`, but strips extras so `"Ur"`->`"r"`, `"rt"`->`"r"`

`"r"` has been the default mode for `open` since at least Py2.4 so retaining it while removing the extra cruft seems like the correct choice.


asottile:
sorry but you're not respecting the boundaries I've put in place for saying no to things

PR #724, opened 2022-10-05
setup.cfg: Add project_urls for PyPI with links to funding, source code and issue tracker pages

EwoutH:
Add [`project_urls`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) to setup.cfg with URLs to the pyupgrade funding page, source code and bug tracker. This will add those URLs under the "Project links" section on https://pypi.org/project/pyupgrade, making it easy to find these resources directly from PyPI.


asottile:
I would rather people land on the repository which is already the case

Issue #532, opened 2021-09-12
Remove inheriting from object when creating classes with type()


verhovsky:
Just as

```py
class A(object): pass
```

is re-written as

```py
class A: pass
```

, so should

```py
A = type("A", (object,), {})
```

be re-written as

```py
A = type("A", (), {})
```

See https://github.com/python/cpython/pull/19553

colin99d:
I am going to attempt this one. @asottile could you assign me?

asottile:
I don't use the assign feature because if someone abandons working on it then it'll look like it's in progress forever and people will be discouraged from even attempting

PR #731, opened 2022-10-08
remove object class from base classes in type calls

lev-blit:
This solves #532


README.md
```diff
@@ -707,3 +707,10 @@ Availability:
 -def f(x: 'queue.Queue[int]') -> C:
 +def f(x: queue.Queue[int]) -> C:
 ```
+
+
+### remove inheriting from object when creating classes with `type()`
+```diff
+-A = type("A", (object,), {})
++A = type("A", (), {})
+```
```
asottile, line 716: can you place this next to the other relevant `object` thing?

tests/features/type_bases_object_test.py
```diff
@@ -0,0 +1,39 @@
...
+        ('A = type("A", (object,), {})', 'A = type("A", (), {})'),
+        ('B = type("B", (object, tuple), {})', 'B = type("B", (tuple,), {})'),
+        (
+            'C = type("C", (object, foo, bar), {})',
+            'C = type("C", (foo, bar), {})',
+        ),
+        ('D = type("D", (tuple, object), {})', 'D = type("D", (tuple,), {})'),
+        (
+            'E = type("E", (foo, bar, object), {})',
+            'E = type("E", (foo, bar), {})',
```
asottile, line 33: I had initially attempted this rewrite as well but ran into all sorts of edge cases where the tuple spans more than one line and/or has trailing commas -- can you include some tests for that case as well

pyupgrade/_token_helpers.py
```diff
@@ -352,6 +352,60 @@ def remove_base_class(i: int, tokens: list[Token]) -> None:
         del tokens[left:last_part + 1]
 
 
+def remove_base_class_from_type_call(i: int, tokens: list[Token]) -> None:
+    # look forward and backward to find commas / parens
```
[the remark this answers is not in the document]
lev-blit: Understood, added arguments to `remove_base_class` to support the difference between the cases

pyupgrade/_plugins/type_bases_object.py
```diff
@@ -0,0 +1,106 @@
...
+        should_move_right_edge=_should_move_right_edge,
+        not_right_edge=_not_right_edge,
+        last_part_function=_last_part_function,
+        do_brace_stack_pop_loop=False,
+        remove_bases=_remove_bases,
+        single_base_char=',',
+        multiple_first_char=')',
+        multiple_first=_multiple_first,
+        multiple_last=_multiple_last,
+    )
```
asottile, line 85: this is a maintenance nightmare, I can't accept the code in this state

lev-blit: well there is a significant difference between `A = type(...)` and `class A(...)` so even though most of the logic is the same, you can see for yourself how much differs(or at least how much differs in my implementation)
this is either something that can be solved by entirely separating the cases(which will cause code duplication) or by someone more experienced / more adapt at this sort of thing tackling the issue.

do you think this is something that I should keep working on, even though it may take a few more tries(if I even find a better solution with my current skill set), or should we close the PR here and leave this for someone more experienced to tackle?

asottile: > well there is a significant difference between A = type(...) and class A(...)

then that should be a dead giveaway that they should **NOT** share code

lev-blit: well that is what I thought at first but you requested that they do(unless I misunderstood your comment as to which function that is)

"this function should reuse the function call parsing function rather than duplicating its work"

lev-blit: well I agreed with that, but it seems I have misunderstood a comment of yours, understanding that you wanted them to share code(said comment - "this function should reuse the function call parsing function rather than duplicating its work")
😞

lev-blit: thank you and sorry for the misunderstanding, will use it :)

lev-blit:
@asottile what should be done for `(foo, object, bar)` and similar cases where object isn't first or last? This doesn't seem like something anyone should write, but maybe it should be supported(or on the contrary, explicitly unsupported)

asottile:
so the weird thing is `type('a', (), {})` in python 2.x also makes a new-style class so it also seems like something someone shouldn't write too:

```pycon
>>> x = type('a', (), {})
>>> x.__bases__
(<type 'object'>,)
```

I would probably remove it anyway 🤷

pyupgrade/_plugins/type_bases_object.py
```diff
@@ -0,0 +1,86 @@
...
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_open_paren
+from pyupgrade._token_helpers import parse_call_args
+
+
+def remove_base_class_from_type_call(i: int, tokens: list[Token]) -> None:
+    type_start = find_open_paren(tokens, 0)
+    bases_start = find_open_paren(tokens, type_start + 1)
+    bases, end = parse_call_args(tokens, bases_start)
+    for base_start, base_end in bases:  # pragma: no cover
```
asottile, line 22: can't ship with this -- means you haven't tested everything

this function is way too complicated -- I don't think it's correct either.  you should also be able to use the original `object` base index without needing to re-find it here and utilize the argument removal functions

pyupgrade/_plugins/type_bases_object.py
```diff
@@ -0,0 +1,86 @@
...
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_open_paren
+from pyupgrade._token_helpers import parse_call_args
+
+
+def remove_base_class_from_type_call(i: int, tokens: list[Token]) -> None:
+    type_start = find_open_paren(tokens, 0)
+    bases_start = find_open_paren(tokens, type_start + 1)
```
asottile, line 20: this isn't necessarily correct -- `type(class_name(), (...))`

pyupgrade/_plugins/type_bases_object.py
```diff
@@ -0,0 +1,86 @@
...
+                    if len(bases) == 2:
+                        # remove the trailing comma,
+                        # there is one after the first base
+                        if tokens[base_end].src == ',':
+                            del tokens[base_start:base_end + 1]
+                        else:
+                            del tokens[base_start:base_end]
+                    # (foo, bar, ..., object[,]) -> (foo, bar, ...[,])
+                    else:
+                        if tokens[base_end - 1].name == UNIMPORTANT_WS and \
```
asottile, line 57: don't use backslashes

lev-blit:
I won't have much time to work on this anymore :/ -- closing the PR so others might take a stab at this

p.s. thank you for being patient with me through this (my first contribution), this was an educating experience :)

Issue #714, opened 2022-10-03
removing "r" in open(..., "r") is not an upgrade


nmichlo:
While `open("foo", "r")` may be considered redundant, some may arguably consider it **explicit** instead, since It is not always immediately obvious that `open("foo")` defaults to read only without first consulting the python documentation.

I do, however, agree that the following is an appropriate change for pyupgrade, since the `U` flag is explicitly deprecated.
```
-open("foo", "rUb")
+open("foo", "rb")
```

The "r" flag on the other hand is not deprecated, rather it when used it is often for readability purposes, and so removing it is a stylistic choice, not necessarily an upgrade.

asottile:
feel free to not use the opinionated tool if you don't agree with the opinions it implements

PR #743, opened 2022-10-28
Enable an option to disable open mode

julien-lecomte:
This merge request adds the "--disable-redundant-open-mode" argument as an opt-in option for this pre-commit hook.

It disables the open_mode part of the plugin.

My team, myself included, would like to keep the redudant "r" read flag when using open.


rkm:
FYI #714

julien-lecomte:
Is that a hard no?

It's an opt-in.

Issue #567, opened 2021-11-15
Do not upgrade string annotations types wrapped in `Optional` to new Union | formatting


wakemaster39:
Simplistic python code:

```python
qqq: "int" | None = 0
```

Output:

```
python qqq.py
Traceback (most recent call last):
  File "qqq.py", line 2, in <module>
    qqq: "int" | None = 0
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

When running pyupgrade a segment of code was updated to use the `|` notation and caused this explosion. I worked around it, but I am sure others might run into similar issues.

asottile:
you've either opted into a python version that's newer than the one you're running or something else you're not showing -- pyupgrade only does this if it's safe

wakemaster39:
@asottile i am running latest python 3.10

asottile:
well then there's something you're not showing me -- watch this video and try again: https://youtu.be/ritp4gAqNMI

wakemaster39:
you are right,  I didn't finish the bug report. Sorry mac pressed submit. Not use to a github repo getting such quick attention, which is awesome btw. Here is the rest you need to replicate the issue fully.

```python
from typing import Optional

from sqlalchemy.orm import Mapped, relationship

qqq: Mapped[Optional["WorkspaceModel"]]
```

After running pyupgrade, the output I am getting is:

```
from sqlalchemy.orm import Mapped

qqq: Mapped["WorkspaceModel" | None]
```

asottile:
pyupgrade does not change that for for me, please watch the video and follow what it asks you to do

wakemaster39:
1. I am running pyupgrade using pre-commit hooks to keep my code based with the latest syntax
2. I have the following in my pre-commit hooks file:

```
  - repo: https://github.com/asottile/pyupgrade
    rev: v2.29.0
    hooks:
      - id: pyupgrade
        args:
          - --py310-plus
```

When running a `pre-commt run --all` on a file (qqq.py) with a file containing the content of:

```python
from typing import Optional

from sqlalchemy.orm import Mapped

qqq: Mapped[Optional["WorkspaceModel"]]
```

3. I would expect the file to have no changes written to it.

4. What is happening is that a file with the following content is written:

```python
from sqlalchemy.orm import Mapped

qqq: Mapped["WorkspaceModel" | None]
````

As an additional note, the output does not change if I use `pyupgrade qqq.py` instead of using the pre-commit hook.

[...]

PR #750, opened 2022-11-09
Fix `Optional["ForwardRef"]` rewriting

mhils:
`Optional["ForwardRef"]` currently requires two `pyupgrade` calls to be fully rewritten. Here's a minimal example:

```shell
$ echo "from __future__ import annotations; from typing import Optional; x: Optional['str']" > repro.py

$ pyupgrade --py39-plus repro.py; cat repro.py
Rewriting repro.py
from __future__ import annotations; from typing import Optional; x: Optional[str]

$ pyupgrade --py39-plus repro.py; cat repro.py
Rewriting repro.py
from __future__ import annotations; from typing import Optional; x: str | None
```

The problem here is that `typing_pep604.py` bails if it encounters a forward annotation (https://github.com/asottile/pyupgrade/issues/567). This PR adjusts the check to guess ahead if annotations will be dequoted by `typing_pep563.py`. 

I guess there's an argument to be made that forward refs and `from __future__ import annotations` in the same file aren't best practice to begin with, but I've encountered that in real-world code and figured I'd send a fix. No hard feelings if you consider this a wontfix. :)

### DIFF
--- a/pyupgrade/_plugins/typing_pep604.py
+++ b/pyupgrade/_plugins/typing_pep604.py
@@ -144,16 +144,17 @@ def visit_Subscript(
     if not _supported_version(state):
         return
 
-    # prevent rewriting forward annotations
-    if (
-            (sys.version_info >= (3, 9) and _any_arg_is_str(node.slice)) or
-            (
-                sys.version_info < (3, 9) and
-                isinstance(node.slice, ast.Index) and
-                _any_arg_is_str(node.slice.value)
-            )
-    ):
-        return
+    # don't rewrite forward annotations (unless we know they will be dequoted)
+    if 'annotations' not in state.from_imports['__future__']:
+        if (
+                (sys.version_info >= (3, 9) and _any_arg_is_str(node.slice)) or
+                (
+                    sys.version_info < (3, 9) and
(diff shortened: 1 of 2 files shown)


pyupgrade/_plugins/typing_pep604.py
```diff
@@ -144,14 +144,18 @@ def visit_Subscript(
...
+    )
     if (
-            (sys.version_info >= (3, 9) and _any_arg_is_str(node.slice)) or
-            (
-                sys.version_info < (3, 9) and
-                isinstance(node.slice, ast.Index) and
-                _any_arg_is_str(node.slice.value)
-            )
+        has_forward_annotation and
+        'annotations' not in state.from_imports['__future__']
```
asottile, line 158: while it probably doesn't actually matter in practice -- the `'annotations'` in check is very cheap and should probably come before the computation for forward annotations

mhils: Good point, fixed. I've kept the `has_forward_annotation` variable for readability, but that is subjective. Let me know if you want me to make it a single big if statement. :)

asottile (approved):

PR #759, opened 2022-11-24
unittest aliases: check for single positional argument

### LINKED ISSUE #749: unittest.makeSuite() with prefix incorrectly rewritten
```py
return unittest.makeSuite(Test, 'test')
```

is rewritten to:

```py
return unittest.defaultTestLoader.loadTestsFromTestCase(Test, 'test')
```

which is invalid:

```
    return unittest.defaultTestLoader.loadTestsFromTestCase(Test, 'test')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: TestLoader.loadTestsFromTestCase() takes 2 positional arguments but 3 were given
```

This works:
```py
return unittest.defaultTestLoader.loadTestsFromTestCase(Test)
```

sstiglitz:
Closes #749

### DIFF
--- a/pyupgrade/_plugins/unittest_aliases.py
+++ b/pyupgrade/_plugins/unittest_aliases.py
@@ -59,7 +59,9 @@ def visit_Call(
             isinstance(node.func, ast.Attribute) and
             isinstance(node.func.value, ast.Name) and
             node.func.value.id == 'unittest' and
-            node.func.attr in FUNCTION_MAPPING
+            node.func.attr in FUNCTION_MAPPING and
+            len(node.args) == 1 and
+            len(node.keywords) == 0
     ):
         func = functools.partial(
             replace_name,
--- a/tests/features/unittest_aliases_test.py
+++ b/tests/features/unittest_aliases_test.py
@@ -7,7 +7,7 @@ from pyupgrade._main import _fix_plugins
 
 
 @pytest.mark.parametrize(
-    ('s',),
+    's',
     (
         pytest.param(
             'class ExampleTests:\n'
@@ -15,6 +15,13 @@ from pyupgrade._main import _fix_plugins
             '        self.assertEqual(1, 1)\n',
             id='not a deprecated alias',
         ),
+        # https://github.com/asottile/pyupgrade/issues/749
+        'unittest.findTestCases(Tests, "arg")',
+        'unittest.findTestCases(Tests, prefix="arg")',
+        'unittest.makeSuite(Tests, "arg")',
+        'unittest.makeSuite(Tests, prefix="arg")',
+        'unittest.getTestCaseNames(Tests, "arg")',
+        'unittest.getTestCaseNames(Tests, prefix="arg")',
     ),
 )
 def test_fix_unittest_aliases_noop(s):


tests/features/unittest_aliases_test.py
```diff
@@ -15,6 +15,11 @@
             '        self.assertEqual(1, 1)\n',
             id='not a deprecated alias',
         ),
+        # https://github.com/asottile/pyupgrade/issues/749
+        ('unittest.findTestCases(MyTests, "pos_arg")',),
+        ('unittest.findTestCases(MyTests, "pos_arg", "pos_arg_again")',),
+        ('unittest.makeSuite(MyTests, "pos_arg")',),
+        ('unittest.getTestCaseNames(MyTests, "pos_arg")',),
```
asottile, line 22: if they're only 1-arg I don't think you need tuples

tests/features/unittest_aliases_test.py
```diff
@@ -64,6 +69,21 @@ def test_fix_unittest_aliases(s, expected):
...
+        (
+            'unittest.makeSuite(MyTests, kw="arg")',
+            'unittest.defaultTestLoader.loadTestsFromTestCase' +
+            '(MyTests, kw="arg")',
+        ),
+        (
+            'unittest.getTestCaseNames(MyTests, kw="arg")',
+            'unittest.defaultTestLoader.getTestCaseNames' +
+            '(MyTests, kw="arg")',
+        ),
```
asottile, line 86: I think these are broken too -- the only valid rewrite case is exactly 1 positional argument

sstiglitz: Sorry, my mistake. Yes, that's the only valid scenario. I updated the logic and the test cases.

pyupgrade/_plugins/unittest_aliases.py
```diff
@@ -59,7 +59,9 @@ def visit_Call(
             isinstance(node.func, ast.Attribute) and
             isinstance(node.func.value, ast.Name) and
             node.func.value.id == 'unittest' and
-            node.func.attr in FUNCTION_MAPPING
+            node.func.attr in FUNCTION_MAPPING and
+            len(node.args) == 1 and
+            len(node.keywords) == 0
```
asottile, line 64: there's a helper for star args as well

tests/features/unittest_aliases_test.py
```diff
@@ -7,14 +7,21 @@
...
             '        self.assertEqual(1, 1)\n',
             id='not a deprecated alias',
         ),
+        # https://github.com/asottile/pyupgrade/issues/749
+        'unittest.findTestCases(Tests, "arg")',
+        'unittest.findTestCases(Tests, prefix="arg")',
+        'unittest.makeSuite(Tests, "arg")',
+        'unittest.makeSuite(Tests, prefix="arg")',
+        'unittest.getTestCaseNames(Tests, "arg")',
+        'unittest.getTestCaseNames(Tests, prefix="arg")',
```
asottile, line 24: don't need a test for each of them since there's no special logic between them.  just the minimal to demonstrate the change

PR #788, opened 2023-01-27
Add option for auto-detecting running python as rewrite target

DjLegolas:


asottile:
the discussion on the issue hasn't come to a decision on the name yet -- please contribute to that, and I don't like your choice of name either :P

PR #790, opened 2023-02-01
Update README.md about unsupported string

Gustry:
Great tool, thanks.

However, I add to look in the test file https://github.com/asottile/pyupgrade/blob/286156f1050244a3c2a459eb56a9dd9223b12833/tests/features/percent_format_test.py#L145
to understand if it's a bug or a feature request.

So I think adding a note in the readme would be welcome.

Maybe, a workaround would be to add a flag saying that I have checked the code manually, and all remaining ` 'city : %s' % city` can be replaced safely with the `format` function ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ Availability:
 +'{a} {b}'.format(a=1, b=2)
 ```
 
+`'%s' % a` is not supported because the tool can't detect if `a` is a tuple or not.
+
 ### Unicode literals
 
 ```diff


asottile:
no thanks

Gustry:
Ok, thanks for the quick response...

I spent nearly 30 minutes trying to understand why this tool was not working. So I was just trying to help people having the same issue and scratching their heads.

So for other people finding this PR, you can use another tool https://github.com/ikamensh/flynt

asottile:
my issue tracker is not an advertising platform. especially for abusive behaviors

PR #792, opened 2023-02-02
feat: Add --check-only for performing linting only (w/o changing files)

mblo:
A small PR to add `--check-only` mode that does not change any file. This makes it easier to use `pyupgrade` during a linting stage.


asottile:
please be a good open source citizen and search the issue tracker

PR #825, opened 2023-05-13
fix #819, TimoutError aliases

### LINKED ISSUE #819: replace socket.timeout on py3.10+ and asyncio.TimeoutError/concurrent.futures.TimeoutError on py3.11+ with TimeoutError
There are now a number of deprecated aliases for `TimeoutError`:

https://docs.python.org/3/library/socket.html#socket.timeout

> Changed in version 3.10: This class was made an alias of [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError).

https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.TimeoutError

> Changed in version 3.11: This class was made an alias of [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError).

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.TimeoutError

> Changed in version 3.11: This class was made an alias of [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError).

lancelote:
* socket.timeout 3.10+
* asyncio.TimeoutError 3.11+
* concurrent.futures.TimeoutError 3.11+

### DIFF
--- a/pyupgrade/_plugins/timeout_error_aliases.py
+++ b/pyupgrade/_plugins/timeout_error_aliases.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+import ast
+import functools
+from typing import Iterable
+
+from tokenize_rt import Offset
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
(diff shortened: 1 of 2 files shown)


asottile:
there's a little more complexity here -- you'll want to take a look at how `oserror_aliases` is implemented since there's special handling for the `except` blocks

PR #840, opened 2023-06-12
Add --check-only and --diff options

ThomasFaivre:
Hi all,

I would like to be able to visualize what modifications would be applied by pyupgrade without modifying files on disk.

These are options that are available in other linting tools like black, isort, autoflake, and others.

What do you think?

I am hesitating between --check and --check-only because we see both in other tools.
I would also like to see some colors on the diff (I like colors! :D), but that would require either some external library (which can be annoying) or handling terminal color codes in pyupgrade directly (which can be annoying aswell). I can be a separate PR though!

Last thing, would it be possible to have a version 3.3.3 with these patches so that it can be available for Python 3.7? I don't have much ground to stand on for this as it is the version for Debian 10 which is oldoldstable since Saturday ^^
Your call!

Thanks!

### DIFF
--- a/pyupgrade/_main.py
+++ b/pyupgrade/_main.py
@@ -329,9 +329,12 @@ def _fix_file(filename: str, args: argparse.Namespace) -> int:
     if filename == '-':
         print(contents_text, end='')
     elif contents_text != contents_text_orig:
-        print(f'Rewriting {filename}', file=sys.stderr)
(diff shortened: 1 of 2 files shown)


asottile:
general open source advice:

(1) search the issue tracker for duplicates
(2) start with an issue, especially for large feature work

PR #875, opened 2023-08-19
Add usage example to README.md

Flimm:
Before this change, there is no indication in the `README.md` that this tool can be run on the command-line. This change makes it clear in `README.md` that this PyPI package can be run on the command-line.


asottile:
stop

PR #885, opened 2023-09-12
Unify exception replacing and rewrite some TimeoutError cases

mxr:
Relates to #819. Moved `oserror_aliases` into `exceptions` & made it generalizable. Added `ExceptionRewriteTarget` object to store module & builtin rewrite config.


pyupgrade/_plugins/exceptions.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import annotations
```
asottile, line 1: I would probably just put all the exception rewriting here -- rather than making importable helpers for the other modules.  should simplify the structure a bit (and make this module less confusing since it's not actually a plugin)

pyupgrade/_plugins/timeouterror_aliases.py
```diff
@@ -0,0 +1,38 @@
...
+
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._plugins import exceptions
+
+EXC_REWRITE_TARGETS = (
+    exceptions.ExceptionRewriteTarget('socket', 'timeout', (3, 10)),
+    exceptions.ExceptionRewriteTarget('asyncio', 'TimeoutError', (3, 11)),
+)
```
asottile, line 16: oh I meant that these can all just live in the `exceptions` plugin -- no need to separate these

asottile:
I rewrote things a bit -- it was doing a bunch of repeated O(N*M) work and didn't work for multiple names being rewritten at once

I also removed all the formatting and magic from the testsuite -- it made it a lot harder to understand what was actually being tested

asottile (approved):

mxr:
Awesome!

PR #901, opened 2023-10-02
Defaultdict lambdas

### LINKED ISSUE #899: Upgrade `defaultdict` default_factory (simple lambdas)
Hello!

It is quite common to see lambda expression used as `default_factory` in  `defaultdict`, however sometime there are some shorter / more idiomatic options, especially for builtin types. Some exemples:

```diff
-defaultdict(lambda: set())
+defaultdict(set)

-defaultdict(lambda: [])
+defaultdict(list)
-defaultdict(lambda: list())
+defaultdict(list)

-defaultdict(lambda: {})
+defaultdict(dict)

-defaultdict(lambda: 0)
+defaultdict(int)

-defaultdict(lambda: 0.0)
+defaultdict(float)

# Doesn't make much sense to use a tuple here thought
-defaultdict(lambda: ())
+defaultdict(tuple)
```
I was wondering if that kind of fixer would be in scope for pyupgrade ?

Checking open source python code, the lambda pattern seems quite common so I think this fixer could benefit many people: 
https://sourcegraph.com/search?q=context%3Aglobal+language%3Apython+defaultdict%5C%28%28lambda%3A+.*%29%5C%29&patternType=regexp&sm=1&groupBy=group&expanded=

If you find it's worth implementing, I'll be happy to work on it.

Thanks

UnknownPlatypus:
Fixes #899 

I left a few comments with some questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,32 @@ Sample `.pre-commit-config.yaml`:
 +{a: b for a, b in y}
 ```
 
+### Replace unnecessary lambdas in `collections.defaultdict` calls
+
+```diff
+-defaultdict(lambda: [])
++defaultdict(list)
+-defaultdict(lambda: list())
++defaultdict(list)
+-defaultdict(lambda: {})
++defaultdict(dict)
+-defaultdict(lambda: dict())
++defaultdict(dict)
+-defaultdict(lambda: ())
++defaultdict(tuple)
+-defaultdict(lambda: tuple())
++defaultdict(tuple)
+-defaultdict(lambda: set())
++defaultdict(set)
+-defaultdict(lambda: 0)
++defaultdict(int)
+-defaultdict(lambda: 0.0)
++defaultdict(float)
+-defaultdict(lambda: 0j)
++defaultdict(complex)
+-defaultdict(lambda: '')
++defaultdict(str)
+```
 
 ### Format Specifiers
 
--- a/pyupgrade/_plugins/defauldict_lambda.py
+++ b/pyupgrade/_plugins/defauldict_lambda.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+import ast
+import functools
+from typing import Iterable
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._ast_helpers import is_name_attr
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_op
+from pyupgrade._token_helpers import parse_call_args
+
+
+def _eligible_lambda_replacement(lambda_expr: ast.Lambda) -> str | None:
+    if isinstance(lambda_expr.body, ast.Constant):
+        if lambda_expr.body.value == 0:
+            return type(lambda_expr.body.value).__name__
+        elif lambda_expr.body.value == '':
+            return 'str'
+        else:
+            return None
+    elif isinstance(lambda_expr.body, ast.List) and not lambda_expr.body.elts:
+        return 'list'
+    elif isinstance(lambda_expr.body, ast.Tuple) and not lambda_expr.body.elts:
+        return 'tuple'
+    elif isinstance(lambda_expr.body, ast.Dict) and not lambda_expr.body.keys:
+        return 'dict'
+    elif (
+            isinstance(lambda_expr.body, ast.Call) and
+            isinstance(lambda_expr.body.func, ast.Name) and
+            not lambda_expr.body.args and
+            not lambda_expr.body.keywords and
+            lambda_expr.body.func.id in {'dict', 'list', 'set', 'tuple'}
+    ):
+        return lambda_expr.body.func.id
+    else:
+        return None
+
+
+def _fix_defaultdict_first_arg(
+        i: int,
+        tokens: list[Token],
+        *,
+        replacement: str,
+) -> None:
+    start = find_op(tokens, i, '(')
+    func_args, end = parse_call_args(tokens, start)
+
+    tokens[slice(*func_args[0])] = [Token('CODE', replacement)]
+
+
+@register(ast.Call)
+def visit_Call(
+        state: State,
+        node: ast.Call,
+        parent: ast.AST,
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+            is_name_attr(
+                node.func,
+                state.from_imports,
+                ('collections',),
+                ('defaultdict',),
+            ) and
+            node.args and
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -751,3 +751,32 @@ Availability:
 -datetime.timezone.utc
 +datetime.UTC
 ```
+
+### Replace unnecessary lambdas in `collections.defaultdict` calls
```
UnknownPlatypus, line 755: Wasn't sure where to put the documentation section so I placed it at the end, let me know if you prefer me to move it somewhere else. (Maybe so exemple could be trimmed down too)

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,77 @@
...
+from pyupgrade._ast_helpers import is_name_attr
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_and_replace_call
+
+_ITERABLE_BUILTINS = frozenset(('dict', 'list', 'set', 'tuple'))
+
+
+def _lambda_constant_replacement(value: Any) -> str | None:
```
UnknownPlatypus, line 20: Should this be inlined in the next function ?

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,77 @@
...
+    if (
+        is_name_attr(
+            node.func,
+            state.from_imports,
+            ('collections',),
+            ('defaultdict',),
+        ) and
+        node.args and
+        isinstance(node.args[0], ast.Lambda) and
+        (replacement := _eligible_lambda_replacement(node.args[0])) is not None
```
UnknownPlatypus, line 65: I haven't seen any walrus usage in pyupgrade but it was convenient here, let me know if you want me to change it thought.

asottile: please don't use walrus -- I would never

README.md
asottile: I would maybe put it after the dict / set comprehension stuff since it's ~related-ish

pyupgrade/_plugins/defauldict_lambda.py
asottile: yeah especially given the use of `Any` which really shouldn't be happening at all!

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _lambda_constant_replacement(value: Any) -> str | None:
+    if value == 0:
+        return 'int' if isinstance(value, int) else 'float'
+    if value is None:
+        return ''
+    if value == '':
+        return 'str'
+    return None
```
asottile, line 27: these should be `elif` + `else` and not repeated `if`s -- same feedback below

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,80 @@
...
+
+        prefix = ''
+        if (
+            isinstance(node.func, ast.Attribute) and
+            isinstance(node.func.value, ast.Name)
+        ):
+            prefix = f'{node.func.value.id}.'
+        func = functools.partial(
+            find_and_replace_call,
+            template=f'{prefix}defaultdict({replacement})',
```
asottile, line 78: this will break calls like `defaultdict(lambda: [], {1: []})`

asottile: it might make sense to parse the function arguments directly and replace it with the arg slice -- rather than going through the full function replacement (which is a bit clunky especially when you need to do the `prefix` bit -- it's more designed for changing the function name than fixing parameters)

you can see something similar to this in the exceptions fixer

tests/features/defauldict_lambda_test.py
```diff
@@ -0,0 +1,217 @@
...
+            'collections.defaultdict(tuple)\n'
+            'collections.defaultdict()\n'
+            'collections.defaultdict(list)\n'
+            'collections.defaultdict(dict)\n'
+            'collections.defaultdict(str)\n'
+            'collections.defaultdict(str)\n'
+            'collections.defaultdict(int)\n'
+            'collections.defaultdict(float)\n'
+            'collections.defaultdict(float)\n'
+            '',
```
asottile, line 210: why a weird empty string there?

pyupgrade/_plugins/defauldict_lambda.py
UnknownPlatypus: Yeah, the prefix bit was really a hack. Your idea is much better here

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,88 @@
...
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._ast_helpers import is_name_attr
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import find_op
+from pyupgrade._token_helpers import parse_call_args
+
+_ITERABLE_BUILTINS = frozenset(('dict', 'list', 'set', 'tuple'))
```
asottile, line 18: this constant is only used once so it's not really useful as a constant

README.md
```diff
@@ -56,6 +56,34 @@ Sample `.pre-commit-config.yaml`:
...
+-defaultdict(lambda: 0)
++defaultdict(int)
+-defaultdict(lambda: 0.0)
++defaultdict(float)
+-defaultdict(lambda: "")
++defaultdict(str)
+-defaultdict(lambda: '')
++defaultdict(str)
+-defaultdict(lambda: None)
++defaultdict()
```
asottile, line 85: this one actually isn't equivalent

pyupgrade/_plugins/defauldict_lambda.py
```diff
@@ -0,0 +1,86 @@
...
+from pyupgrade._token_helpers import parse_call_args
+
+
+def _eligible_lambda_replacement(lambda_expr: ast.Lambda) -> str | None:
+    if isinstance(lambda_expr.body, ast.Constant):
+        if lambda_expr.body.value == 0:
+            if isinstance(lambda_expr.body.value, int):
+                return 'int'
+            else:
+                return 'float'
```
asottile, line 25: there's also `0j` which is `complex`

asottile (commented):
I'm just going to fix these but a heads up to you

adamchainz:
Nice feature, it just fixed seven cases on a client code base. Thanks both!

UnknownPlatypus:
Thanks Adam and thanks Anthony for the thorough review and tweaks!

PR #940, opened 2024-04-02
Add config file support

### LINKED ISSUE #511: Use pyproject.toml as alternative to command line args
### Feature request

Instead of end-user running `pyupgrade --py37-plus`, we can utilise tools section in pyproject.toml like:

```toml
[tool.pyupgrade]
py37-plus = true
```
and just run the command `pyupgrade`

Implementation requirements: Usual requirements for finding pyproject.toml in parents folders if current directory is within the python project/packages applies. Single VCS repository may contain multiple python packages too.

### Motivation
We have `pyupgrade` wired into `pre-commit` hook, but the hook itself is optional in large legacy code as it introduces huge diff. Without pre-commit hook, the onus falls back on developer to remember arguments for each projects or peek into `.pre-commit-config.yaml`.

`isort` has a similar functionality, that after following in `pyproject.toml`
```toml
[tool.isort]
profile = "black"
```
developer can simply run `isort` instead of `isort --profile black`.

chrisbouchard:
Add support for configuration files in pyupgrade using [`ConfigArgParse`][ConfigArgParse-pypi].

[ConfigArgParse-pypi]: https://pypi.python.org/pypi/ConfigArgParse

By default, we'll look for a `tool.pyupgrade` section in `pyupgrade.toml` or `pyproject.toml` (in that order), and read the configuration from there. Command line arguments will override configuration file settings, which will override command line argument defaults. We also add a `--config-file` option to override the config file path.

Fixes #511.

### Example Output

```console
% pyupgrade -h
usage: pyupgrade [-h] [--config-file CONFIG_FILE] [--exit-zero-even-if-changed] [--keep-percent-format] [--keep-mock] [--keep-runtime-typing] [--py3-plus]
                 [--py36-plus] [--py37-plus] [--py38-plus] [--py39-plus] [--py310-plus] [--py311-plus] [--py312-plus]
                 [filenames ...]

positional arguments:
  filenames

options:
  -h, --help            show this help message and exit
  --config-file CONFIG_FILE
                        config file path
  --exit-zero-even-if-changed
  --keep-percent-format
  --keep-mock
  --keep-runtime-typing
  --py3-plus, --py3-only
  --py36-plus
  --py37-plus
  --py38-plus
  --py39-plus
  --py310-plus
  --py311-plus
  --py312-plus

Args that start with '--' can also be set in a config file (pyupgrade.toml or pyproject.toml or specified via --config-file). Config file syntax is Tom's
Obvious, Minimal Language. See https://github.com/toml-lang/toml/blob/v0.5.0/README.md for details. In general, command-line values override config file
values which override defaults.
```

### TODO

* [x] Add `ConfigArgParse` dependency
* [x] Switch from `argparse` to `configargparse`
* [ ] Add unit tests for config file parsing
* [ ] Add documentation for config file handling

### DIFF
--- a/pyupgrade/_main.py
+++ b/pyupgrade/_main.py
@@ -341,7 +341,11 @@ def _fix_file(filename: str, args: argparse.Namespace) -> int:
 
 
 def _build_config_arg_parser() -> configargparse.ArgumentParser:
-    parser = configargparse.ArgumentParser()
+    config_parser = configargparse.TomlConfigParser(['tool.pyupgrade'])
+    parser = configargparse.ArgumentParser(
+        default_config_files=['pyupgrade.toml', 'pyproject.toml'],
+        config_file_parser_class=config_parser,
+    )
 
     parser.add_argument('filenames', nargs='*')
     parser.add_argument('--exit-zero-even-if-changed', action='store_true')
--- a/pyupgrade/_main.py
+++ b/pyupgrade/_main.py
@@ -8,6 +8,7 @@ import tokenize
 from typing import Match
 from typing import Sequence
 
+import configargparse
 from tokenize_rt import NON_CODING_TOKENS
 from tokenize_rt import parse_string_literal
 from tokenize_rt import reversed_enumerate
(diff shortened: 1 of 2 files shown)


asottile:
no thanks, this has been rejected before please search next time

asottile:
oh you did find the duplicate and decided to waste both of our time anyway

PR #948, opened 2024-06-08
rewrite TypeVar defaults for Generator / AsyncGenerator

asottile:


asottile:
once mypy 1.11 lands I will merge this \o/

PR #956, opened 2024-07-29
Add `--py313-plus` flag

hugovk:
Add the flag for PEP 696 `TypeVar` rewrites (added in https://github.com/asottile/pyupgrade/pull/948):

https://github.com/asottile/pyupgrade#pep-696-typevar-defaults


asottile:
not until rc1

PR #966, opened 2024-09-28
Add --py313-plus option

cdce8p:
#948 added a check which conditionally depends on `--py313-plus`. At it as an argument option.


asottile:
this intentionally does not get added until 3.13 final

Issue #970, opened 2024-10-10
Versioned branch fixer removes comments after `if` block


adamchainz:
The 'versioned branch' fixer currently removes comments after a dropped `if` block:

```diff
 import sys
 
-if sys.version_info < (3, 2):
-    print("3.2")
-
-# let's go
 print("ok")
```

The comment is outside of the block and should be preserved.

(Just fixed a similar issue in django-upgrade: https://github.com/adamchainz/django-upgrade/issues/495 .)

asottile:
there's a duplicate somewhere. this is relatively unfixable because there's no definitive end of a block

adamchainz:
Isn't this sufficient? #971 . It seems to, at least.

BonaFideIT:
This could be the duplicate issue which @asottile mentioned: https://github.com/asottile/pyupgrade/issues/931.

But I think @adamchainz is right, @asottile can you recheck this please?

asottile:
there's a case that this creates a syntax error, I can't remember the exact one

PR #971, opened 2024-10-10
Fix removal of comments after dropped version block

adamchainz:
Attempt at fixing #970.


BonaFideIT:
Looks good to me, seams to be the same solution as used in https://github.com/adamchainz/django-upgrade/pull/496/ for django-upgrade.

PR #983, opened 2024-10-24
pep585 import aliases

stephenfin:
Track aliases so that pep585 rewriting works with e.g.

```
import typing
x: ty.List[str] = []
```

### DIFF
--- a/pyupgrade/_plugins/typing_pep585.py
+++ b/pyupgrade/_plugins/typing_pep585.py
@@ -36,7 +36,7 @@ def visit_Attribute(
     if (
             _should_rewrite(state) and
             isinstance(node.value, ast.Name) and
-            node.value.id == 'typing' and
+            state.as_imports.get(node.value.id) == 'typing' and
             node.attr in PEP585_BUILTINS
     ):
         func = functools.partial(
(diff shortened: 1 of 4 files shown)


pyupgrade/_plugins/typing_pep585.py
```diff
@@ -36,7 +36,7 @@ def visit_Attribute(
     if (
             _should_rewrite(state) and
             isinstance(node.value, ast.Name) and
-            node.value.id == 'typing' and
+            state.as_imports.get(node.value.id) == 'typing' and
```
stephenfin, line 39: Note that this breaks cases like we see in `typing_pep604_test.py` below, where we use `typing` without first importing it. I think that's okay since that code is already broken in real-life, but if we wanted to avoid making those changes we could trivially change this to:

```suggestion
        state.as_imports.get(node.value.id, node.value.id) == 'typing'
```

asottile:
aliases present all sorts of problems and so they are intentionally not supported.

stephenfin:
> they are intentionally not supported.

I did my due diligence and couldn't find anything in the code or existing issues to suggest this. Additionally, I see support for aliased objects in multiple plugins. If there's a serious issue I've missed here than a little more context would have been a good learning experience for both me and anyone else that stumbles upon this issue. But fine :shrug: As always, been a pleasure interacting with you, Anthony 🙄

asottile:
the only plug-in with support is the imports one because the result is unambiguous

Issue #982, opened 2024-10-23
PEP 604 typing duplication removal


walter9388:
This is a bit niche, but when I was reviewing some naive code submitted to me I saw some bizarre typing generated by `pyupgrade` which looked something like
```python
def f() -> int | None | None:
    ...
```
where `None` is duplicated. 

When looking into how this had happened, I noticed it came from the following:
```diff
-def f() -> Optional[Union[int, None]]):
+def f() -> int | None | None:
     ...
```

Clearly wrapping `Union[..., None]` with `Optional` is bad, but it made me wonder which dev tool can detect this sort of issue. It doesn't seem to be something that is picked up by `mypy`, `flake8`, `pyright`, etc. 
_(Just to clarify this bad typing had been generated by an internal tool which wraps given types with `Optional` for PATCH methods in a RESTful API, so it is very much an edge case)_

Do you think deduplication of types is something that can live in the PEP604 part of pyupgrade? I think it is unlikely that anyone would write `int | None | None` manually, but I can see when over complicating unions this sort of thing might happen.

I added the following test to `test_fix_pep604_types` just to check this was expected behaviour:
```python
        pytest.param(
            'from typing import Optional, Union\n'
            'def f(x: Optional[Union[int, None]]): pass\n'
            'def g(x: Union[Union[int, None], None]): pass\n'
            'def h(x: Union[int, int, None]): pass\n'
            'def k(x: Union[Union[int, None], int]): pass\n',

            'from typing import Optional, Union\n'
            'def f(x: int | None): pass\n'
            'def g(x: int | None): pass\n'
            'def h(x: int | None): pass\n'
            'def k(x: int | None): pass\n',

            id='nested unions or optionals with duplicated typing',
        ),
```

and sure enough pytest failed with the following duplicated types:
```
    from typing import Optional, Union
  - def f(x: int | None): pass
  + def f(x: int | None | None): pass
  ?               +++++++
  - def g(x: int | None): pass
  + def g(x: int | None | None): pass
  ?               +++++++
  - def ag(x: int | None): pass
  + def ag(x: int | int | None): pass
  ?                ++++++
  - def h(x: int | None): pass
  + def h(x: int | None | int): pass
  ?                    ++++++
```

I'd be interested to hear your opinions on this. If you do think it would be a complimentary feature for `pyupgrade` then I'm happy to submit a PR for consideration to get the test above passing.

asottile:
[...]

PR #1004, opened 2025-02-20
add deduplication of types

walter9388:
Relating to https://github.com/asottile/pyupgrade/issues/982

Took me a while to get round to this, but here we are....

I think there are three different levels of approaching this issue:
1. Remove extra `None` types only:
    ```diff
    -def f(x: Optional[Union[int, None]]): pass
    +def f(x: int | None): pass
    -def g(x: Union[Optional[int], None]): pass
    +def g(x: int | None): pass
    ```
2. Remove any duplicated scalar types at the same depth by name in `Union` blocks:
    ```diff
    -def f(x: Union[Union[Union[Union[a, b], c], d], a]): pass
    +def f(x: a | b | c | d): pass
    -def g(x: Union[a.b | a.c, a.b, list[str], str]): pass
    +def g(x: a.b | a.c | list[str] | str): pass
    ```
3. General deduplication at any depth on any block:
    ```diff
    -def f(x: Union[list[Union[int, str]], list[Union[str, int]]]): pass
    +def f(x: list[int | str]): pass
    ```

I settled on level 2 as this was still possible with a single pass and seemed more useful than just focusing on `None` types.
I couldn't see a way of approaching the general problem (level 3) without recursively making a tree structure and then assessing the leaf nodes. However, I am not deeply familiar with the standard python libraries for parsing ASTs etc., so if there are simple built in methods for problems like this I would be interested to know!

I used the existing scan in `_fix_union` to determine the delimitators at `depth==1` between the types. This seemed to work well, but I definitely ran into some interesting edge cases when it came to handling comments, whitespace and multilines.

I have managed to get this working for a variety of test cases, and I would be interested to hear your feedback.

Btw I enjoy your YouTube content! I have learnt a lot of niche things I would have struggled to pick up otherwise. So thank you for that!

### DIFF
--- a/pyupgrade/_plugins/typing_pep604.py
+++ b/pyupgrade/_plugins/typing_pep604.py
@@ -116,7 +116,9 @@ def _fix_union(
     else:
         comma_positions = []
 
-    to_delete += _find_duplicated_types(tokens, j, top_level_breaks, lines_with_comments)
+    to_delete += _find_duplicated_types(
+        tokens, j, top_level_breaks, lines_with_comments,
+    )
 
     if tokens[j].line == tokens[k].line:
         del tokens[k]
@@ -141,7 +143,10 @@ def _fix_union(
         del tokens[i:j]
 
 
-def _find_closing_bracket_and_if_contains_none(tokens: list[Token], i: int) -> tuple[int, bool]:
+def _find_closing_bracket_and_if_contains_none(
+    tokens: list[Token],
+    i: int,
+) -> tuple[int, bool]:
     assert tokens[i].src in _OPENING
     depth = 1
     i += 1
--- a/tests/features/typing_pep604_test.py
+++ b/tests/features/typing_pep604_test.py
@@ -312,13 +312,15 @@ def f(x: int | str) -> None: ...
         ),
         pytest.param(
             'from typing import Union\n'
-            'def f(x: Union[list[Union[int, str]], list[Union[str, int]]]): pass\n',
+            'def f(x: Union[list[Union[int, str]], list[Union[str, int]]]):\n'
+            '    pass\n',
 
             'from typing import Union\n'
-            'def f(x: list[int | str]): pass\n',
+            'def f(x: list[int | str]):\n'
+            '    pass\n',
 
             id='general duplicated types',
-            marks=pytest.mark.xfail(reason='requires recursive type searching'),
+            marks=pytest.mark.xfail(reason='requires deeper type evaluation'),
         ),
     ),
 )


tests/features/typing_pep604_test.py
```diff
@@ -242,6 +242,84 @@ def f(x: int | str) -> None: ...
...
+            ')\n',
+
+            id='duplicated types in multi-line nested unions or optionals',
+        ),
+        pytest.param(
+            'from typing import Union\n'
+            'def f(x: Union[list[int, str], list[Union[str, int], int]]): pass\n',
+
+            'from typing import Union\n'
+            'def f(x: list[int, str]): pass\n',
```
asottile, line 318: `list[int, str]` isn't a valid type -- was this test meaning to show something else?

pyupgrade/_token_helpers.py
```diff
@@ -166,6 +182,37 @@ def find_block_start(tokens: list[Token], i: int) -> int:
     return i
 
 
+def find_duplicated_types(
```
asottile, line 185: if this is only used in one file it shouldn't live in the general shared file (same for the other function)

asottile (commented):
just a quick first pass -- will look more closely later

pyupgrade/_token_helpers.py
walter9388: Yeah fair. I've moved them both to the `typing_604.py` file.

tests/features/typing_pep604_test.py
walter9388: That was dumb of me. I had it marked as `xfail` so I didn't pick up on using `,` over `|`. I've updated that example now and I'll update the PR message as well.

pyupgrade/_plugins/typing_pep604.py
```diff
@@ -14,22 +14,36 @@
...
+from pyupgrade._token_helpers import _OPENING
 from pyupgrade._token_helpers import find_op
 from pyupgrade._token_helpers import is_close
 from pyupgrade._token_helpers import is_open
 
 
 def _fix_optional(i: int, tokens: list[Token]) -> None:
     j = find_op(tokens, i, '[')
-    k = find_closing_bracket(tokens, j)
+    k, contains_none = _find_closing_bracket_and_if_contains_none(tokens, j)
```
walter9388, line 25: Modified the general `find_closing_bracket` function to also check for whether the optional block already contains `None` in the same pass.

pyupgrade/_plugins/typing_pep604.py
```diff
@@ -14,22 +14,36 @@
...
     if tokens[j].line == tokens[k].line:
-        tokens[k] = Token('CODE', ' | None')
+        if contains_none:
+            del tokens[k]
+        else:
+            tokens[k:k + 1] = [
+                Token('UNIMPORTANT_WS', ' '),
+                Token('CODE', '| '),
+                Token('CODE', 'None'),
+            ]
```
walter9388, line 34: The reason for changing the single token containing ` | None` to explicit whitespace, `|` and `None` is for the deduplication and whitespace removal functions used in `_fix_union`. This also applies to the multiline version a few lines below.

pyupgrade/_plugins/typing_pep604.py
```diff
@@ -110,11 +134,80 @@ def _fix_union(
 
         for comma in comma_positions:
             tokens[comma] = Token('CODE', ' |')
+        to_delete += _remove_consecutive_unimportant_ws(
+            tokens, [x for x in range(j, k) if x not in to_delete],
+        )
```
walter9388, line 139: Not convinced this is the best approach to remove whitespace, but not sure about what to do in situations where lines are completely deleted other than comments. I have written a niche test for this situation in test `id='duplicated types in multi-line nested unions or optionals'`.

walter9388 (commented):
Just highlighting a few things to be aware of.

pyupgrade/_plugins/typing_pep604.py
asottile: it might be helpful to do all the tokens then -- `|` would be `'OP'` and `None` would be `'NAME'`

pyupgrade/_plugins/typing_pep604.py
```diff
@@ -110,11 +134,80 @@ def _fix_union(
...
+        i = d1c
+    return to_delete
+
+
+def _remove_consecutive_unimportant_ws(
+    tokens: list[Token], idxs: list[int],
+) -> list[int]:
+    to_delete = []
+    prev_name = ''
+    for kk in idxs:
```
asottile, line 201: why kk?

PR #1009, opened 2025-04-26
Convert asyncio.coroutine to async def

AtomsForPeace:
I've encountered some old python libraries that still use asyncio.coroutine instead of async def. From what I see this change was around 3.5. I've just added a failing test with xfail on it for now as a start. I'm not even sure this is a relevant feature. It's just something that has annoyed me a couple of times.

### DIFF
--- a/tests/features/asyncio_coroutine_test.py
+++ b/tests/features/asyncio_coroutine_test.py
@@ -1,0 +1,24 @@
+import pytest
+
+from pyupgrade._data import Settings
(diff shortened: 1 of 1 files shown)


mxr:
Seems like these are not 1:1 based on https://stackoverflow.com/a/40573425. Not a maintainer but in my experience, pyupgrade doesn't do rewrites unless it is highly confident the old and new code is the same. Maybe the rewrite could happen in high-confidence situations that are determined through parsing type annotations

Issue #872, opened 2023-08-13
nested context managers


asottile:
`--py39-plus` (though it was only specced for 3.10 -- double check what pypy does here)

```python
with a:
   with b:
        ...
```

to

```python
with (
    a,
    b,
):
    ...
```

slightlybelowzen:
I didn't quite understand what you meant by 'double check what pypy does here'?

asottile:
what don't you understand? what are you confused about?

PR #1012, opened 2025-05-08
Fold nested context managers.

shaleh:
Python 3.10 allows for context managers to be grouped in the same was a imports. Previous versions only worked with a hack: with (foo as _, bar as _):
    body

This PR implements two fixes:
1. Fold nested with statements.

```
with foo:
    with bar:
        body
```

becomes

```
with (foo, bar):
    body
```

2. Removes the pre 3.10 hack

```
with (foo as _, bar as _):
    body
```

becomes

```
with (foo, bar):
    body
```

Resolves. https://github.com/asottile/pyupgrade/issues/872

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,7 +783,6 @@ Availability:
 +def f(x: queue.Queue[int]) -> C:
 ```
 
-
 ### use `datetime.UTC` alias
 
 Availability:
@@ -794,4 +793,24 @@ Availability:
 
 -datetime.timezone.utc
 +datetime.UTC
+```
+
+### Fold nested context managers
+
+Availability:
+- `--py310-plus` and higher
+
+```diff
+- with foo:
+-     with bar:
+-         body
++ with (foo, bar):
++     body
+```
+
+```diff
+- with (foo as _, bar as _):
+-     body
++ with (foo, bar):
++     body
 ```
--- a/pyupgrade/_plugins/fold_nested_context_managers.py
+++ b/pyupgrade/_plugins/fold_nested_context_managers.py
@@ -1,0 +1,136 @@
+from __future__ import annotations
+
+import ast
+import functools
+import itertools
+from collections.abc import Iterable
+from typing import Any
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from pyupgrade._ast_helpers import ast_to_offset
+from pyupgrade._data import register
+from pyupgrade._data import State
+from pyupgrade._data import TokenFunc
+from pyupgrade._token_helpers import Block
+
+
+def _expand_item(indent: int, item: ast.AST) -> str:
+    return '{}{}'.format(' ' * indent, ast.unparse(item))
+
+
+def _replace_context_managers(
+    i: int,
+    tokens: list[Token],
+    *,
+    with_items: list[ast.withitem],
+    body: Iterable[ast.AST],
+) -> None:
+    block = Block.find(tokens, i, trim_end=True)
+    block_indent = block._minimum_indent(tokens)
+    replacement = '{}with ({}):\n{}\n'.format(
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -795,3 +794,23 @@ Availability:
...
+```diff
+- with foo:
+-     with bar:
+-         body
++ with (foo, bar):
++     body
+```
+
+```diff
+- with (foo as _, bar as _):
```
asottile, line 812: I don't think this is in scope.  there's no reason to write the original code with nonsense as

shaleh: This was a real work around. I know, I maintain code with it. So I added code to remove it.

asottile: no it does not.  nobody would write code like that because it doesn't help

shaleh: https://docs.python.org/3.3/reference/compound_stmts.html#the-with-statement

You needed to use the `as` syntax to stack them in one `with`. But since you often did not need the name using `_` was a real thing.

The options were:
```
with foo:
    with bar:
```
or
```
with foo as f, bar as b:
```

As you correctly point out, the underscore names are not helpful and with 3.10 there is nothing forcing the programmer to use a name. So it is ok to remove them. Technically.... we could do analysis of the body and following code to see if any of the names are used but that gets ugly fast so I stuck with only removing the useless underscore.

asottile: you don't need the as even in 3.10

asottile: right and I am telling you it is out of scope for pyupgrade.  it's not a language syntactic improvement

shaleh: Disagree but understood. Will post a revised version shortly.

shaleh: In case we misunderstood each other.

3.3 to 3.9 did require the use of `as` to nest context managers. `with foo as f, bar as b:` was the only way.
3.10 remove the requirement for the `as` naming. Now you can do `with (foo, bar):` and it follows the same pattern as import statements.

My original PR removed the now unneeded syntax of the `as` when migrating to 3.10 when the name was an underscore because that is was semantically safe. Any other name may have actually been used. I am moving this logic to another branch because I still want to upgrade code that I have to maintain elsewhere. I can post it as a separate PR for clarity. Or it can remain in a private repo. No worries.

shaleh:
PR updated as requested. In doing so I found a case I had fixed and then somehow missed during subsequent debugging.

shaleh:
I would appreciate insight into why the 3.9 code coverage is failing. It passes for 3.10.

README.md
asottile: no it did not.  I just tried it and it works fine without as

the only thing 3.9 added was trailing commas

shaleh: Hmm, our code was 3.7 at one point. I wonder if this was changed between somewhere before 3.9. Because I know it was failing CI or we would not have added the ugly captures. Either way, it is out of the current PR. Folding the nesting is still valuable.

shaleh: Ok, I did some digging. Looks like that weird underscore form came from a StackOverflow discovery back in like 3.3 or 3.5 days.... definitely worthless now.

This code below passes 3.9 but the parenthesis form at the end does not parse in 3.8 or older. The `as` syntax does not help.

Thanks for your patience. I have removed one more mislearned thing from my list.

```
from contextlib import suppress

try:
    with suppress(ValueError) as _, suppress(IndexError) as _:
        print("hello 1")
        raise ValueError("foo")
except ValueError:
    print("Not supressed")

try:
    with suppress(ValueError), suppress(IndexError):
        print("hello 2")
        raise ValueError("foo")
except ValueError:
    print("Not supressed")

try:
    with (
        suppress(ValueError),
        suppress(IndexError)
    ):
        print("hello 3")
        raise ValueError("foo")
except ValueError:
    print("Not supressed")
```

PR #1018, opened 2025-06-08
Document "Python2 and old Python3.x blocks" case that causes `SyntaxError`

mxr:


asottile:
I think the test cases are good enough -- I don't think this needs to be documented explicitly beyond the line that's already there

PR #1056, opened 2026-04-03
Rewrite `async_timeout`

mxr:


pyupgrade/_plugins/imports.py
```diff
@@ -196,6 +196,8 @@
         ('typing_extensions', 'is_typeddict'): 'typing',
     },
     (3, 11): {
+        ('async_timeout', 'timeout'): 'asyncio',
+        ('async_timeout', 'timeout_at'): 'asyncio',
```
asottile, line 200: these are generated, you'll want to edit the source instead

mxr: Oops https://github.com/asottile/reorder-python-imports/pull/433

mxr:
Not a safe rewrite, see https://github.com/asottile/reorder-python-imports/pull/433 for discussion

PR #1066, opened 2026-05-18
rewrite concurrent.futures.TimeoutError to TimeoutError on py311+

mokashang:
Closes the remaining piece of #819: rewrite `concurrent.futures.TimeoutError` to the builtin `TimeoutError` when `--py311-plus` (or later) is in effect.

In Python 3.11 `concurrent.futures.TimeoutError` became an alias for the builtin `TimeoutError`, the same way `asyncio.TimeoutError` did. #885 already handles `socket.timeout` (3.10+) and `asyncio.TimeoutError` (3.11+); this PR adds the third member of that set.

### Approach

* Add `_Target('TimeoutError', 'concurrent.futures', 'TimeoutError', (3, 11))` to `_TARGETS` in `pyupgrade/_plugins/exceptions.py`.
* Add `'concurrent.futures'` to `RECORD_FROM_IMPORTS` so a `from concurrent.futures import TimeoutError` is tracked the same way `asyncio` is.
* The `ast.Attribute` branch of `_get_rewrite` previously only handled single-segment module names (it required `node.value` to be an `ast.Name`). I extracted a small `_is_dotted_name` helper that walks an `Attribute` chain so dotted modules like `concurrent.futures` are matched too. The helper degenerates to the original `Name` check for single-segment modules, so the existing `asyncio.TimeoutError`/`os.error`/etc. rewrites continue to behave identically.

### Behavior

```python
# input (py311+)
raise concurrent.futures.TimeoutError(1)
try: ...
except concurrent.futures.TimeoutError: ...
try: ...
except (concurrent.futures.TimeoutError, ValueError): ...

# output
raise TimeoutError(1)
try: ...
except TimeoutError: ...
try: ...
except (TimeoutError, ValueError): ...
```

On py310 and below the rewrite is a noop, matching the existing behavior for `asyncio.TimeoutError`.

### Tests

Added to `tests/features/exceptions_test.py`:

* `test_fix_exceptions_version_specific_noop` cases for `raise`/`except concurrent.futures.TimeoutError` at `(3, 10)`.
* `test_fix_exceptions_versioned` cases for the `raise`, `except`, and `except (...,)` rewrites at `(3, 11)`.
* `test_can_rewrite_dotted_module_alongside_others` covering the interaction with the existing disparate-names tuple rewrite.

Full test suite: `1017 passed, 4 xfailed`. `pre-commit run --all-files` clean (flake8, autopep8, mypy).

### DIFF
--- a/pyupgrade/_data.py
+++ b/pyupgrade/_data.py
@@ -39,6 +39,7 @@ RECORD_FROM_IMPORTS = frozenset((
     'asyncio',
     'collections',
     'collections.abc',
+    'concurrent.futures',
     'functools',
     'mmap',
     'os',
--- a/pyupgrade/_plugins/exceptions.py
+++ b/pyupgrade/_plugins/exceptions.py
@@ -36,7 +36,17 @@ _TARGETS = (
     _Target('OSError', None, 'WindowsError', (3,)),
     _Target('TimeoutError', 'socket', 'timeout', (3, 10)),
     _Target('TimeoutError', 'asyncio', 'TimeoutError', (3, 11)),
+    _Target('TimeoutError', 'concurrent.futures', 'TimeoutError', (3, 11)),
 )
+
+
+def _is_dotted_name(node: ast.AST, dotted: str) -> bool:
+    parts = dotted.split('.')
+    for part in reversed(parts[1:]):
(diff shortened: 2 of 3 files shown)


asottile:
we do not accept ai slop contributions

Issue #1071, opened 2026-07-17
pyupgrade reverts TypedDict from typing_extensions to typing, breaks PEP-728 types


baluyotraf:
## Background

[PEP-728] added the ability to provide the `closed` and `extra_items` keywords in 3.15. Now there is a reason to `typing_extensions`.

[PEP-728]: https://peps.python.org/pep-0728/

## Reproduce

**upgrade.py**
```python
from typing_extensions import TypedDict

class MyClass(TypedDict, closed=True):
  ...
```

**run**
```
pyupgrade --py313-plus --keep-runtime-typing upgrade.py
```

## Result

**upgrade.py**
```python
from typing import TypedDict

class MyClass(TypedDict, closed=True):
  ...
```

**run**
```
python upgrade.py
Traceback (most recent call last):
  File "/home/raf/dbtune/research-experimentation/upgrade.py", line 4, in <module>
    class MyClass(TypedDict, closed=True):
      ...
TypeError: _TypedDictMeta.__new__() got an unexpected keyword argument 'closed'
```

## Workaround

For now I simply imported `typing_extensions` directly and used `typing_extensions.TypedDict`. (I did alias it to make it shorter).

asottile:
find the place where it's generated from and send a patch

PR #1072, opened 2026-07-17
fix: remove the 3.13 TypedDict replacement to accomodate PEP-728

baluyotraf:
In relation to #1071 

The replacement will probably come back for 3.15 but at the moment there is no 3.15 code.

The tests only test the general replacement logic and I'm assuming that these definitions here
are not tested on its own, so I didn't change the test files.


asottile:
lol why do I even put comments in the file

PR #1081, opened 2026-08-16
`argparse_suggest_on_error`: fix missing container for a string

Andrej730:
Issue was mostly harmless, but was making plugin apply on weird cases like:

```python
import argparse

argparse.ArgumentPars(suggest_on_error=True)
```

`is_name_attr` expecting `Container[str]` and technically just `str` also passes as a container of strings.

https://github.com/asottile/pyupgrade/blob/44e8a53fed5f7a8a43765180f9515349050739ce/pyupgrade/_ast_helpers.py#L22-L27

### DIFF
--- a/pyupgrade/_plugins/argparse_suggest_on_error.py
+++ b/pyupgrade/_plugins/argparse_suggest_on_error.py
@@ -27,7 +27,7 @@ def visit_Call(
                 node.func,
                 state.from_imports,
                 ('argparse',),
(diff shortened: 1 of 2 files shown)


asottile:
write a test (it'll still fail lol you haven't actually fixed it)

Andrej730:
Ahh, sorry :face_with_peeking_eye:
Fixed it

asottile (approved):

PR #1082, opened 2026-08-17
identity_equality: fix ints-floats bug/support complex and tuple

Andrej730:
Hi! Fixing 1 bug in `identity_equality` and expanding fix to cover 2 more types. 

Possibly could have been 3 PRs, but changes are interleaved, so submitting them as one.
Commits are separated for readability, but I'll summarize them here also.

1. For skipping bools check was relying on `in {True, False}`:

https://github.com/asottile/pyupgrade/blob/f08b199e76297c88173992813df2cd00614b9601/pyupgrade/_plugins/identity_equality.py#L38-L40

Since `in` is using `==` as a fallback, so the check end up also skipping `1.0`, `0.0`, `0`, `1` and cases like `x is 1`, etc.

2. Added support for `complex` numbers, they also have a warning at runtime. Though use of this is minimal is probably.

```
>>> a = 5j
>>> a is 5j
<python-input-2>:1: SyntaxWarning: "is" with 'complex' literal. Did you mean "=="?
```

3. Added support for tuples that have all its elements as constants (including singletons that are usually skipped), Python warns about them too - they're also optimized during Python compilation as constants and sometimes `x is (1,2,3)` results to `True`, relying on the implementation detail.
In theory we can also warn about nested tuples, but they're never optimized by Python and hence it never warns about them, so it's a different kind of error - same as if it would be with `is []`, `is {}`, etc.

```python
>>> a is (1,2,3)
<python-input-5>:1: SyntaxWarning: "is" with 'tuple' literal. Did you mean "=="?
False
>>> a is (None, True, ...)
<python-input-1>:1: SyntaxWarning: "is" with 'tuple' literal. Did you mean "=="?
False
>>> a is ((1,2,3), 4)
False
>>> a is ([1,2,3], 4)
False
```

### DIFF
--- a/tests/features/identity_equality_test.py
+++ b/tests/features/identity_equality_test.py
@@ -35,6 +35,9 @@ def test_fix_is_compare_to_literal_noop(s):
         pytest.param('x is u""', 'x == u""', id='unicode string'),
         pytest.param('x is b""', 'x == b""', id='bytes'),
         pytest.param('x is 1.5', 'x == 1.5', id='float'),
+        # Regression tests - ensure we don't mistake those cases for `x is True`.
+        pytest.param('x is 1', 'x == 1', id='int equals bool True'),
+        pytest.param('x is 1.0', 'x == 1.0', id='float equals bool True'),
         pytest.param('x == 5 is 5', 'x == 5 == 5', id='compound compare'),
         pytest.param(
             'if (\n'
--- a/tests/features/identity_equality_test.py
+++ b/tests/features/identity_equality_test.py
@@ -44,7 +44,7 @@ def test_fix_is_compare_to_literal_noop(s):
         pytest.param('x is (None,)', 'x == (None,)', id='tuple with None'),
         pytest.param('x is (True, 1)', 'x == (True, 1)', id='tuple with bool'),
         pytest.param('x is (...,)', 'x == (...,)', id='tuple with ellipsis'),
-        # Regression tests - ensure we don't mistake those cases for `x is True`.
+        # Regression tests - ensure we don't mistake the cases for `x is True`.
         pytest.param('x is 1', 'x == 1', id='int equals bool True'),
         pytest.param('x is 1.0', 'x == 1.0', id='float equals bool True'),
         pytest.param('x == 5 is 5', 'x == 5 == 5', id='compound compare'),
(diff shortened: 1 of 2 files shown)


tests/features/identity_equality_test.py
```diff
@@ -12,14 +12,17 @@
         'x is True',
         'x is False',
         'x is None',
+        'x is ...',
         'x is (not 5)',
         'x is 5 + 5',
-        # pyupgrade is timid about containers since the original can be
+        # pyupgrade is timid about mutable containers since the original can be
```
asottile, line 18: rewriting the comment doesn't make it incorrect -- the tests are there for a very good reason

```pycon
>>> x = (1, 2, 3)
>>> x is (1, 2, 3)
<stdin>:1: SyntaxWarning: "is" with 'tuple' literal. Did you mean "=="?
False
>>> x == (1, 2, 3)
True
>>> 
```

asottile:
this patch does way too much.  this should be separated into reviewable (and correct ideally) parts

tests/features/identity_equality_test.py
Andrej730: The same applies to all literals, tuples are no special.

```python
>>> a = 1234
>>> a is 1234
<python-input-1>:1: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
False
>>> a == 1234
True
```

asottile: not all literals

```pycon
>>> x = 5
>>> x is 5
<stdin>:1: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
True
>>> x == 5
True
```

Andrej730: Yes, there are some cases when it's always-true, but the check don't differentiate and allows all ints, floats, strs, bytes.
So technically it's unsafe changing `is` for `==` for any of them, unless we bake into the check exact implementation details when Python does reuse literals - e.g. first couple hundreds ints you've mentioned. Not sure what else can be considered absolutely safe without deep analysis of where the comparison occurs.

Andrej730:
Commits are splitted to make them easier to follow, but I'm happy to split this into 3 PRs.

But it seems we need to decide on `tuple` first, since it's more fundamental issue for this plugin - if `tuple` can't be supported, because it doesn't have any instances where it would be always-True, then floats, bytes, partially ints/strings should be excluded too.

PR #1087, opened 2026-08-21
Rewrite `__aiter__`/`__anext__` in Python 3.10+

### LINKED ISSUE #960: Convert `__anext__` calls on Python 3.10+
Python 3.10 introduces the [`anext`](https://docs.python.org/3/library/functions.html#anext) builtin which calls `__anext__` on the provided async iterator. It would be great if `pyupgrade` could automatically convert legacy `async_iterator.__anext__()` calls into `anext(async_iterator)` upon upgrading to Python 3.10 or above.

Andrej730:
Hi! Adding a plugin that converts `__aiter__()`/`__anext__()` dunder attributes calls to `aiter`/`anext`.

Mostly, everything is straightforward, except this nested call - since `__anext__` is visited first, it generates a callback and it gets applied first, `__aiter__` applied second. And it's a problem, since they both share the same starting token index. So previously this would result in `aiter(anext(a))`, which is incorrect.

https://github.com/asottile/pyupgrade/blob/ef0d20d5d0d6bafaca75192176d3d87adc29262b/tests/features/async_dunders_test.py#L57-L61

So to correct this, I've changed `pyupgrade` to run callbacks in the reversed order, so more nested ast elements would be fixed first. 

https://github.com/asottile/pyupgrade/blob/ef0d20d5d0d6bafaca75192176d3d87adc29262b/pyupgrade/_main.py#L79-L82

The only other place required to be changed is `subprocess_run` plugin - its `universal_newlines` fix relies on `stderr`/`stdout` arg removed beforehand. There's an existing "both universal_newlines and capture_output rewrite" test that was able to catch this issue:

https://github.com/asottile/pyupgrade/blob/b9df209584a5b40f01730932cac01927cd81210e/tests/features/capture_output_test.py#L101-L106

Resolves https://github.com/asottile/pyupgrade/issues/960

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,6 +748,20 @@ Availability:
      ...
 ```
 
+### rewrite `__aiter__`/`__anext__` in Python 3.10+
+
+Availability:
+- `--py310-plus` is passed on the commandline.
+
+```diff
+-x.__aiter__()
++aiter(x)
+```
+
+```diff
+-x.__anext__()
++anext(x)
+```
 
 ### pep 604 typing rewrites
 
@@ -873,19 +887,4 @@ Availability:
 ```diff
 -parser = argparse.ArgumentParser(suggest_on_error=True)
 +parser = argparse.ArgumentParser()
-```
-
-### rewrite `__aiter__`/`__anext__` in Python 3.10+
-
-Availability:
-- `--py310-plus` is passed on the commandline.
-
-```diff
--x.__aiter__()
-+aiter(x)
-```
-
-```diff
--x.__anext__()
-+anext(x)
 ```
--- a/README.md
+++ b/README.md
@@ -874,3 +874,18 @@ Availability:
 -parser = argparse.ArgumentParser(suggest_on_error=True)
 +parser = argparse.ArgumentParser()
 ```
+
+### rewrite `__aiter__`/`__anext__` in Python 3.10+
+
+Availability:
+- `--py310-plus` is passed on the commandline.
+
+```diff
+-x.__aiter__()
++aiter(x)
+```
+
+```diff
+-x.__anext__()
++anext(x)
+```
--- a/pyupgrade/_main.py
+++ b/pyupgrade/_main.py
@@ -76,9 +76,7 @@ def _fix_plugins(contents_text: str, settings: Settings) -> str:
             continue
         # though this is a defaultdict, by using `.get()` this function's
         # self time is almost 50% faster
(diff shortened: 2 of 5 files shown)


pyupgrade/_main.py
```diff
@@ -76,7 +76,9 @@ def _fix_plugins(contents_text: str, settings: Settings) -> str:
             continue
         # though this is a defaultdict, by using `.get()` this function's
         # self time is almost 50% faster
-        for callback in callbacks.get(token.offset, ()):
+        # Reversed order is needed to properly handle to the nested attrs.
+        # E.g. in `a.xxx.yyy`, `.xxx` to get fix applied before `.yyy`.
+        for callback in reversed(callbacks.get(token.offset, ())):
```
asottile: rather than change this just fix your new plugin that doesn't follow what the others do

README.md
```diff
@@ -874,3 +874,18 @@ Availability:
 -parser = argparse.ArgumentParser(suggest_on_error=True)
 +parser = argparse.ArgumentParser()
 ```
+
+### rewrite `__aiter__`/`__anext__` in Python 3.10+
```
asottile: this should go roughly by the 3.10ish things

pyupgrade/_main.py
Andrej730: The problem is that then we need to special case `a.xxx`, lookup its parent, since it will be already handled during `a.xxx.yyy` visit. So reversing the order of callbacks, making it consistent with ast nesting, just makes things simpler. It's similar problem to why `_fix_plugins` is using `reversed_enumerate` for tokens in general.

asottile: then it shouldn't be coupled to this PR.  please stop putting a bunch of changes into PRs and split them into separate ones so they're more easily reviewable and revertable

Andrej730: But it was needed for PR to work.

Anyway, reworked PR, dropped other modules changes in a6ef3c5 and made `async_dunders` plugin work without the general pyupgrade changes in 2733304

README.md
Andrej730: e17872f - moved it to appear after py39 fixes and before py310 typing rewrites fix

pyupgrade/_main.py
asottile: I mean it's maybe fine to improve here -- but it should be a separate PR

Andrej730: This part is removed now, pr has no changes outside the new plugin code.

PR #1102, opened 2026-09-15
exceptions: handle unparenthesized exception tuples

### LINKED ISSUE #1096: Python 3.14 PEP 758 unparenthesized exception rewrites of deprecated exceptions
[PEP 758](https://peps.python.org/pep-0758/) makes the parentheses around multiple exceptions optional when there's no `as` clause. pyupgrade offers a rewrite of `asyncio.TimeoutError` to just `TimeoutError` since https://github.com/asottile/pyupgrade/pull/885. When the two are involved, the output is a valid Python code.

### Minimal reproduction
In my actual production code the `try..except` was in the body of the preceding function but the issue reproduces even when it's at the module level.

#### env
```
$ python -V
Python 3.14.7
$ pip freeze | grep pyupgrade
pyupgrade==3.21.2
```

#### input
```
$ cat repro.py
import asyncio

try:
    pass
except asyncio.TimeoutError, ValueError:
    pass


def f(x) -> None:
    pass
```

#### run
```
$ pyupgrade --py314-plus repro.py
Rewriting repro.py
```

#### result

`TimeoutError` is inserted at the closest `(` which in this case is the next function.
```
$ cat repro.py
import asyncio

try:
    pass
except asyncio.TimeoutError, ValueError:
    pass


def f TimeoutError -> None:
    pass
```

#### expected:
```
import asyncio

try:
    pass
except TimeoutError, ValueError:
    pass


def f(x) -> None:
    pass
```

Rayan-and-beyond:
Fixes #1096.

PEP 758 allows multiple exception types without wrapping parentheses. The exception rewrite callback previously assumed every tuple-shaped handler started with `(`, so it could search forward into unrelated code and rewrite there.

This keeps the existing fully-parenthesized tuple path (including deduplication) and falls back to rewriting individual exception AST offsets when the exception tuple is unparenthesized or only partially parenthesized.

Validation:
- reproduced the reported corruption on current `main` with Python 3.14.7
- exception feature module: 40 passed on Python 3.14
- full Python 3.14 coverage run: 1104 passed / 1 skipped / 4 xfailed, 100% coverage
- full Python 3.11 coverage run: 1099 passed / 4 skipped / 6 xfailed, 100% coverage
- `pre-commit run --all-files --show-diff-on-failure` passed (including flake8 and mypy)


asottile:
your ai slop is not welcome here

